PR #20, opened 2021-12-09
add heading hierarchy

nanxstats:
This PR adds an extra level of hierarchy to the vignette headings (e.g. from h1 to h2) so that the floating TOC can be rendered correctly under > pkgdown 2.0.0.


nanxstats:
@elong0527

elong0527 (approved):

PR #44, opened 2022-10-13
ongoing fix

LittleBeannie:
fix `eAccural`


nanxstats:
I'm merging this because looks like it's urgent. Please do fix any issues this causes later.

nanxstats:
Since this would introduce an additional hard dependency, is `purrr::reduce()` really necessary here? Maybe `base::Reduce` is good enough?

PR #46, opened 2022-10-14
42 gs power ahr has futility bounds when test lower = false

LittleBeannie:


nanxstats (changes requested):
`is.infinite(x)` is probably a more robust choice than `abs(x) != Inf`.

LittleBeannie:
> `is.infinite(x)` is probably a more robust choice than `abs(x) != Inf`.

Agree!

nanxstats:
OK, it looks like you got some merging conflicts in the vignettes - @LittleBeannie please resolve them so that we can merge this.

nanxstats (approved):

PR #55, opened 2022-10-16
update gitignore

elong0527:
- we should keep snapshot testing results.
- ignore `doc` or `docs` to avoid people accidentally upload `pkgdown` results from local.


elong0527:
The error in macOS seems not relevant to this PR.

nanxstats (changes requested):

nanxstats (approved):
I updated this by removing `doc/` from `.Rbuildignore` and `.gitignore`.

PR #56, opened 2022-10-16
move old functions

elong0527:
- move old functions to `inst/old_function`


elong0527:
@LittleBeannie , could you confirm if `gridpts_h1_hupdate_oldR.R` should be moved as well?

LittleBeannie:
> @LittleBeannie , could you confirm if `gridpts_h1_hupdate_oldR.R` should be moved as well?

Yes!

LittleBeannie (approved):

PR #60, opened 2022-10-17
change `info` into `info0` when calculate crossing prob under H0

LittleBeannie:


nanxstats (approved):
It looks like this can be either a bug fix or maybe this change the way the results are generated - my point is, in PRs, it would be good to document the intention and reasoning of the change, instead of only stating what has been done.

Issue #62, opened 2022-10-25
Use fixed bound in `gs_power_bound` gives error


LittleBeannie:
The following application of `gs_power_combo` gives the error message.
```
enrollRates <- tibble::tibble(Stratum = "All", duration = 12, rate = 500/12)

failRates <- tibble::tibble(Stratum = "All",
                            duration = c(4, 100),
                            failRate = log(2) / 15,  # median survival 15 month
                            hr = c(1, .6),
                            dropoutRate = 0.001)

fh_test <- rbind( data.frame(rho = 0,
                             gamma = 0,
                             tau = -1,
                             test = 1,
                             Analysis = 1:3,
                             analysisTimes = c(12, 24, 36)),
                  data.frame(rho = c(0, 0.5),
                             gamma = 0.5,
                             tau = -1,
                             test = 2:3,
                             Analysis = 3,
                             analysisTimes = 36))

gs_power_combo_test1 <- gsDesign2::gs_power_combo(enrollRates = enrollRates, 
                                                  failRates = failRates, 
                                                  fh_test = fh_test, 
                                                  upper = gs_b, upar = c(3, 2, 1), 
                                                  lower = gs_b, lpar = c(-1, 0, 1))
```

```
Error in checkmvArgs(lower = lower, upper = upper, mean = mean, corr = corr, :
‘lower’ not specified or contains NA
```

The root cause of the above error is from `gs_bound`. This function works perfectly for spending bound, but maybe not fixed bound. For example, in `gs_power_combo`, the application of `gs_bound` is 
```
bound <- gs_bound(alpha = upper(upar, min_info_frac),
                    beta = lower(lpar, min_info_frac),
                    analysis = info_fh$Analysis,
                    theta = theta_fh * sqrt(n),
                    corr = corr_fh,
                    binding_lower_bound = binding,
                    algorithm = algorithm,
                    alpha_bound = identical(upper, gs_b),
                    beta_bound = identical(lower, gs_b),
                    ...)
```
If in the fixed-bound case above, we have 
```
bound <- gs_bound(alpha = gs_b(c(3, 2, 1)),
                    beta = gs_b(c(-1, 0, 1)),
                    analysis = c(1, 2, 3, 3, 3),
                    theta = c(0.8916756, 2.6121208, 3.4484207, 3.8160358, 3.7788639),
                    corr = matrix(c( 1.0000000, 0.6614295, 0.5724133, 0.3377230, 0.4128853,
[...]

PR #70, opened 2022-10-28
enlarge tolerance

LittleBeannie:
Let's increase the testing tolerance first as a TEMPORARY solution, so the CMD check can be successfully passed. After issue #71 and #62 are solved, need to re-visit it and decrease the testing tolerance.


nanxstats (approved):
I appreciate simple solutions!

(please do remember to restore it once ready)

PR #76, opened 2022-10-28
remove warning messages

LittleBeannie:


nanxstats (approved):
This patch is very much needed to reduce the noise level in certain cases. I hope they are well-documented in other places such as the table footnotes!

PR #77, opened 2022-10-28
75 re orgclean up of pkgdown vignettes tab

LittleBeannie:


_pkgdown.yml
```diff
@@ -30,122 +30,162 @@ navbar:
...
-  - story_npe_integration
-  - story_arbitrary_distribution
-  - story_compute_expected_events
-  - story_risk_difference
-  - NPH_Futility
-- title: "Use case of functions"
-  navbar: ~
-  desc: >
-    The following vignettes cover the usage of the key functions in gsDesign2.
+    The following vignette is a quick start for you to learn gsDesign quickly!
```
nanxstats, line 110: gsDesign -> gsDesign2

nanxstats: Avoid using `!` in formal and technical/business documentation.

_pkgdown.yml
```diff
@@ -30,122 +30,162 @@ navbar:
...
   - text: Articles
     href: articles/index.html
+  - text: Examples
+    href: articles/index.html
   - text: News
     href: news/index.html
 
 reference:
-- title: High-level Functions
+- title: "High-level Functions"
```
nanxstats, line 39: You should prefer sentence case instead of title case for these "titles" which are actually "headings".

vignettes/usage_fixed_design.Rmd
```diff
@@ -9,7 +9,7 @@ output:
     code_folding: hide
     highlight: "textmate"
     css: "custom.css"
-bibliography: gsDesign.bib
+bibliography: ggsd.bib
```
nanxstats, line 12: Should this be `gsDesign2.bib` instead?

nanxstats (changes requested):
Overall looking good. I added a few comments.

vignettes/usage_fixed_design.Rmd
LittleBeannie: gsDesign.bib and ggsd.bib share a lot of bib in common. Since Keaven initially called it ggsd.bib, I keep it as it is there. Let's get back later after Keaven approves the vignettes list.

_pkgdown.yml
LittleBeannie: Nice suggestion! Can we keep the titles here, since the new pkgdown layout only serves as an initialization for Keaven's review?

nanxstats (approved):

PR #81, opened 2022-10-31
add theta1 and info1

LittleBeannie:


nanxstats (approved):
Hey @LittleBeannie for better documentation and future references, please always write something in the PR description to detail what is changed or fixed and the "why".

PR #96, opened 2022-11-11
84 api

LittleBeannie:


nanxstats (approved):

nanxstats:
Congratulations finishing this "large-scale change". 👍 

I wonder if the input argument order of certain functions such as `expected_accrual()` should be adjusted following the "data, descriptor, details" order. https://design.tidyverse.org/args-data-details.html but that can probably happen in a later PR.

LittleBeannie:
> Congratulations finishing this "large-scale change". 👍
> 
> I wonder if the input argument order of certain functions such as `expected_accrual()` should be adjusted following the "data, descriptor, details" order. https://design.tidyverse.org/args-data-details.html but that can probably happen in a later PR.

Good suggestion! Let's polish it in another PR.

PR #98, opened 2022-11-15
add `r` documentation

LittleBeannie:


nanxstats (changes requested):
This looks good - would you mind addressing two minor details:

1. Keep it one space between r and the description instead of two spaces.
2. Normally r will not be changed by the user. -> Normally, `r` will not be changed by the user.

LittleBeannie:
> This looks good - would you mind addressing two minor details:
> 
> 1. Keep it one space between r and the description instead of two spaces.
> 2. Normally r will not be changed by the user. -> Normally, `r` will not be changed by the user.

Thanks for the suggestion, and updated!

nanxstats (approved):

PR #105, opened 2022-11-26
Add vignette on accuracy of group sequential bounds

nanxstats:
This PR adds a vignette from Keaven on the numerical accuracy of group sequential bounds.

The vignette contains a preliminary examination on some key design parameters and gives a tentative recommendation, adequate to suggesting stopping the practice of reporting more than 6 digits for p-value boundaries. Beside numerical accuracy, this also highlights the importance of a single, clean locked database at the time of analysis before unblinding.

Future extensions to binomial and TTE endpoints would be awesome.

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -147,6 +147,10 @@ articles:
   contents: 
   - starts_with("usage_")
 
+- title: "Best practices"
+  contents:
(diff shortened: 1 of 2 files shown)


keaven (commented):
Thanks for the update. Looks good.

PR #106, opened 2022-11-30
Add Magirr-Burman with maximum up-weighting

elong0527:
Following https://github.com/Merck/simtrial/pull/22#pullrequestreview-1199737602 we add a new weight function `wlr_weight_mb`.

Formula: $$w(t, \tau, w_{\max}) = \min\left(w_{\max},\left(\frac{1}{S(\min(t,\tau))}\right)\right).$$

Below are example code for weight function and sample size calculation

```
enrollRates <- tibble::tibble(Stratum = "All", duration = 12, rate = 500/12)

failRates <- tibble::tibble(Stratum = "All",
                            duration = c(4, 100),
                            failRate = log(2) / 15,  # median survival 15 month
                            hr = c(1, .6),
                            dropoutRate = 0.001)

gs_arm <- gs_create_arm(
  enrollRates,
  failRates,
  ratio = 1, # Randomization ratio
  total_time = 36 # Total study duration
)
arm0 <- gs_arm[["arm0"]]
arm1 <- gs_arm[["arm1"]]


plot(1:36, wlr_weight_mb(x = 1:36, arm0, arm1,tau = 4),
     xlab = "Calendar time (month)", ylab = "Weight", type = "l"
)

plot(1:36, wlr_weight_mb(x = 1:36, arm0, arm1,tau = 4, wmax = 1.1),
     xlab = "Calendar time (month)", ylab = "Weight", type = "l"
)

my_ratio <- 1
my_alpha <- 0.025
my_beta <- 0.2
my_power <- 1 - my_beta

my_analysisTimes <- c(12, 24, 36)

gsDesign2::gs_design_wlr(
  enrollRates = enrollRates,
  failRates = failRates,
  weight = function(x, arm0, arm1){wlr_weight_mb(x, arm0, arm1, tau = 4)},
  analysisTimes = my_analysisTimes,
  alpha = my_alpha,
  beta = my_beta,
  ratio = my_ratio,
  # test.type = 4 non-binding futility bound
  binding = FALSE,
  # alpha spending
  upper = gs_spending_bound,
  upar = list(sf = gsDesign::sfLDOF, total_spend = 0.025),
  # beta spending
  lower = gs_spending_bound,
  lpar = list(sf = gsDesign::sfLDOF, total_spend = 0.2))


gsDesign2::gs_design_wlr(
  enrollRates = enrollRates,
  failRates = failRates,
[...]

PR #107, opened 2022-12-01
update document

elong0527:
update doc per https://github.com/Merck/gsDesign2/pull/106#pullrequestreview-1199934015

### DIFF
--- a/R/wlr_weight.R
+++ b/R/wlr_weight.R
@@ -22,7 +22,7 @@
 #' * `wlr_weight_power` is Gehan-Breslow and Tarone-Ware weight function.
 #' * `wlr_weight_mb` is  Magirr (2021) weight function.
 #' 
-#' @param x analysis time
+#' @param x a vector of numeric value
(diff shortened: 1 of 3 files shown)


nanxstats (changes requested):
First, it should be "numeric value**s**". Second, I would hope the documentation could explain what the argument is representing, instead of only the technical specifications they follow. Does this make sense?

elong0527:
Those weight function are general. I feel it is better to keep the description general as well.

R/wlr_weight.R
```diff
@@ -22,7 +22,7 @@
 #' * `wlr_weight_power` is Gehan-Breslow and Tarone-Ware weight function.
 #' * `wlr_weight_mb` is  Magirr (2021) weight function.
 #' 
-#' @param x analysis time
+#' @param x a vector of numeric value
```
elong0527, line 25:
```suggestion
#' @param x a vector of numeric values
```

PR #111, opened 2022-12-05
`EF` -> `Event fraction`

LittleBeannie:
Add another full name display for event fraction, which is used in the max combo tests.


nanxstats (approved):

nanxstats:
These changes look good! ✨

PR #113, opened 2022-12-07
112 leftover api for rd

LittleBeannie:


nanxstats (approved):
This is a quite challenging task - Personally I wouldn't know how to refactor "N" to "n" because a simple search & replace is obviously not enough... so kudos to you 💯

PR #123, opened 2022-12-15
Fix check notes

nanxstats:
This PR fixes a few significant `R CMD check` notes, such as "no visible binding for global variable".


LittleBeannie (approved):
That's AWESOME checks! Is it ready for a GitHub release now?

PR #134, opened 2023-01-03
update check of failrate

LittleBeannie:


nanxstats (approved):
I can see this is checking:

- Only 1 number in `fail_rate$fail_rate` needs to be > 0.
- All rates should be >=0.

So I'm merging this.

PR #145, opened 2023-01-19
fix lpar bug

LittleBeannie:


yuliasidi (commented):
looks good to me

nanxstats:
> > ```r
> > lpar, rep(-Inf, K)
> > ```
> 
> Since the logic is not 100% equivalent

I'm interested to learn why they are not equivalent - feels like I was simply flipping the conditioning order of both if-else statements. Perhaps I'm missing something?

nanxstats:
@LittleBeannie Do you plan to make any updates? I can merge this if there is no plan to make further changes.

LittleBeannie:
> @LittleBeannie Do you plan to make any updates? I can merge this if there is no plan to make further changes.

No, that's everything I have! But I owe you a counter-example!

nanxstats (approved):
Let's merge this before it's getting merge conflicts.

Issue #140, opened 2023-01-09
Make the styler workflow pass on main


nanxstats:
It looks like the Style (pull_request) run always fails in pull requests, perhaps due to some technical issues in the workflow.

I reproduced the issue independently and submitted an issue to upstream: <https://github.com/r-lib/actions/issues/683>

Let's wait for the feedback from the maintainers.

PR #146, opened 2023-01-20
Update {styler} workflow from upstream to fix the error in PRs

nanxstats:
This PR fixes https://github.com/Merck/gsDesign2/issues/140, as the upstream workflow has been fixed now.


LittleBeannie (approved):
Thanks for the updates!

PR #147, opened 2023-01-20
Accelerate code examples with elapsed time >5 sec

howardbaek:
- Wrapped code examples with elapsed time >5 sec in `\donttest{}` 
- Documented elapsed time of code examples in `inst/example_run_time/timer.Rmd`


nanxstats (approved):
Looking good. Thanks for the patch!

PR #152, opened 2023-02-08
Improve S3 methods documentation

howardbaek:
- Unified the help page of related S3 methods for `as_gt()`, `to_integer()`  and `summary()`
- Got rid of `...` parameter from the function arguments (let's see if the R CMD check throws any warnings or notes)


howardbaek:
Seems like the `R-CMD-check / windows-latest (release) (pull_request)` is failing because of the `...` argument in the functions I changed.

nanxstats (approved):
So it looks like we do need the dot-dot-dot in this case. I further improved the patch to fix the check issues. Thanks for looking into this! I think we learned something from it.

howardbaek:
> So it looks like we do need the dot-dot-dot in this case. I further improved the patch to fix the check issues. Thanks for looking into this! I think we learned something from it.

This was definitely a learning experience. Thanks for the fix!

PR #154, opened 2023-02-10
update rd part

LittleBeannie:


nanxstats (changes requested):
Please consider using `snake_case` for the option values of the enumerated type arguments per the [style guide](https://merck.github.io/gsDesign2/articles/style.html#enumerated-type-arguments).

```r
c("un_stratified", "ss", "invar_h1", "invar_h0")`
```

Also, should it just be `"unstratified"` or `"un_stratified"`? Because the former can be an actual word without a separator.

LittleBeannie:
> Please consider using `snake_case` for the option values of the enumerated type arguments per the [style guide](https://merck.github.io/gsDesign2/articles/style.html#enumerated-type-arguments).
> 
> ```r
> c("un_stratified", "ss", "invar_h1", "invar_h0")`
> ```
> 
> Also, should it just be `"unstratified"` or `"un_stratified"`? Because the former can be an actual word without a separator.

Thanks for reminding me! Yes, let's use `"unstratified"` instead.

LittleBeannie:
It is ready for merge @nanxstats , though there are still some remaining lint issues...

nanxstats (approved):
Now it passes!

PR #159, opened 2023-02-22
Pass linting workflow

nanxstats:
This PR fixes all remaining code linting issues by:

- Fixing the specific issue if it is obvious.
- Adding rules to `.lintr` to exclude specific linters (object name linter and commented code linter) for specific files (mostly tests) if the proper fix is too much overhead.

Now all workflows should pass, including the lintr workflow.


LittleBeannie (approved):
Finally get lintr passed!

PR #165, opened 2023-03-01
decrease tolerance

LittleBeannie:


nanxstats (approved):
Thanks! I further removed the previous comment about the need to decrease the tolerance.

PR #179, opened 2023-03-09
Remove CRAN link for now to pass urlchecker

nanxstats:
This PR removes the non-existent CRAN link for the badge so it passes `urlchecker::url_check()`.


LittleBeannie (approved):
Now the URL test will be smoothly passed!

Issue #22, opened 2022-09-14
Standardize enumerated options naming


nanxstats:
I don't think the naming convention of the enumerated options in function arguments is covered by the current [syntax style guide](https://style.tidyverse.org/syntax.html) for tidyverse, but in a consistent manner with variables/functions, I would propose to use lower case abbreviations and snake case for the enumerated options in function arguments. This means

```r
# Bad
f <- function (method = c("AHR", "WLR", ...) { }

# Good
f <- function (method = c("ahr", "wlr"), ...) { }
```

```r
# Bad
f <- function (approx= c("event driven", "asymptotic", "generalized schoenfeld", ...) { }

# Good
f <- function (approx= c("event_driven", "asymptotic", "generalized_schoenfeld", ...) { }
```

This style is also consistent with tensorflow and keras, and would make your code look consistent in style.

If not done already, we should also make an effort to [enumerate all possible options](https://design.tidyverse.org/def-enum.html) in the function definition, and document them properly.

cc @keaven @elong0527

elong0527:
Agree. Would be good to have an PR to review the impact before we start to implement testing.

keaven:
Thanks for getting us on a good page...I guess we should have our coding conventions on a portal page somewhere or in a vignette in gsDesign2.

nanxstats:
Great, I will send a PR to create a style guide & rules vignette.

There is a good [chapter](https://abseil.io/resources/swe-book/html/ch08.html) about this topic in the SWE book.

PR #182, opened 2023-03-09
Standardize enumerate and class names

nanxstats:
Fixes https://github.com/Merck/gsDesign2/issues/22. I looked into all enumerated options in function arguments and this seems to be the last thing that's not using snake case.

Also, I updated the S3 class name `"non-binding"` to `"non_binding"` to be consistent with all other class names, for example, `gs_design`, `fixed_design`.


LittleBeannie (approved):
Nice catch!

PR #196, opened 2023-03-20
Add installation guide for CRAN version

nanxstats:
This PR completes the installation guide in the README by adding the CRAN installation command.


LittleBeannie (approved):
Thanks for adding the CRAN installation guidance to users!

PR #198, opened 2023-03-21
Fix outdated argument names and use canonical style for text elements

nanxstats:
This PR fixes the outdated argument names and uses canonical style for text elements in the README.


LittleBeannie (approved):
Thanks for your careful check!

PR #204, opened 2023-03-24
update sample size with `info_scale`

LittleBeannie:


nanxstats:
Probably pending what you want to do with the definition of `info_scale`. in #203.

nanxstats (approved):

PR #210, opened 2023-04-12
update `to_integer` and alpha input in `gs_design_ahr`

LittleBeannie:


LittleBeannie:
Hi @nanxstats, please review this PR later tomorrow. I have some new commits working on after talking with Keaven. It will be ready around tomorrow afternoon.

nanxstats (approved):
This looks like a decent improvement on integer size designs. The new vignette looks informative. So I'm going to approve this.

(Although I'm concerned about expressions like `ifelse(condition, FALSE, TRUE)` which is equivalent to `!condition`, these things can probably be simplified in the future.)

PR #211, opened 2023-04-14
update lf in `fixed_design`

LittleBeannie:


nanxstats (approved):
Although I'm a bit worried about the growing cyclomatic complexity of `fixed_design()`, I'm going to approve this as it fixes the bug which produces wrong results.

PR #213, opened 2023-04-18
Fix check notes on missing global function or variable

nanxstats:
This PR fixes the newly introduced check notes on missing global functions or variables.


LittleBeannie (approved):
Thanks for the detailed check!

PR #217, opened 2023-04-21
remove convergence of upper/lower bound in max combo

LittleBeannie:


LittleBeannie:
This will be the last issue before `v1.0.8`.

nanxstats (approved):
As long as there is no new check issues caused by the change, I trust your judgement to remove this logic.

Issue #222, opened 2023-05-15
Format footnote numbers using decimal notation


nanxstats:
At the bottom of the readme, the `as_gt()` output footnote 3 says:

> 3... The smaller value subtracts the probability of crossing a futility bound before crossing an efficacy bound at a later analysis (0.025 - **6e-04** = 0.0244) under the null hypothesis.

We should format 0.025 - **6e-04** = 0.0244 as 0.025 - **0.0006** = 0.0244 and enforce decimal notation for all numbers in the footnotes.

This is because scientific notation might create potential confusion and misinterpretation sometimes. Using decimal notation will ensure clarity and consistency, making it easier for a wider audience including regulators and non-technical stakeholders to understand the information being presented.

LittleBeannie:
Hi @nanxstats , I will not have the bandwidth for this issue in the next 2 months. If possible, could you please find other people from BAAMR for this?

LittleBeannie:
@nanxstats will lead SP members for this ticket. Please feel free to reach out to me if there are any technical questions.

PR #223, opened 2023-05-21
Format alpha in footnote using decimal notation

nanxstats:
This PR fixes https://github.com/Merck/gsDesign2/issues/222

It uses `format(..., scientific = FALSE)` to format the alpha values in the table footnote using decimal notation.

Note that `scientific = FALSE` only solve this formatting issue in general. Since there is this `digits` option (default is 7) in the global options, therefore calling something like

```r
format(0.025 - 6e-10, scientific = FALSE)
```

will give 0.025 instead of 0.024999999.

Although we can set a specific higher `digits` value in the `format()` calls, I choose to leave that option to the users. One can increase the `digits` in the global options or use `withr::with_options()` to change the number of digits displayed:

```r
options(digits = 8)
format(0.025 - 6e-10, scientific = FALSE)
```

This will give 0.024999999.

We can revisit this decision later if this option has to be made more explicit.

### DIFF
--- a/R/as_gt.R
+++ b/R/as_gt.R
@@ -542,15 +542,22 @@ as_gt.gs_design <- function(x,
     x <- x %>%
       gt::tab_footnote(
         footnote = paste0(
-          "Cumulative alpha for final analysis (", x_alpha,
-          ") is less than the full alpha (", full_alpha, ") when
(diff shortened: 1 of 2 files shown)


LittleBeannie (approved):
Thanks, Nan!

PR #225, opened 2023-05-23
Organize Rcpp functions across multiple files

### LINKED ISSUE #224: Split C++ functions into individual cpp and header files
Currently, all C++ functions are organized in a single file `gridpts_h1_hupdate.cpp`. This is suboptimal for scaling up the C++ code base. To improve reusability, we should:

- Split it into different `.cpp` files, one for each function.
- Create an `.h` header files for each `.cpp` file, with "include guard" (`#ifndef`, `#define`, and `#endif`) and the function declaration.
- Use other functions in `.cpp` files with statements like `#include "gridpts.h"`.

nanxstats:
This PR fixes #224 by reorganizing the Rcpp functions across multiple files.

Although only `gridpts` is currently reused by the other two Rcpp functions, I still created a header file for `h1` and `hupdate` to make them future-proof and make the code structure easier to understand at a glance.


elong0527 (approved):
LGTM

nanxstats:
Thanks!

Issue #228, opened 2023-06-06
Rename `foo()`


nanxstats:
There are two internal functions defined inline named `foo()`:

https://github.com/Merck/gsDesign2/blob/0f34f0145321cd1d6606c7a6f992522062ea06ea/R/expected_time.R#L112

https://github.com/Merck/gsDesign2/blob/0f34f0145321cd1d6606c7a6f992522062ea06ea/R/gs_design_combo.R#L195

These functions should be updated:

1. Rename them to use a meaningful function name. See [foo](https://developers.google.com/style/word-list#foo) in Google's developer documentation style guide.
1. Define them outside the export function. This is because currently they are only accessible within the parent function, and it is being redefined again every time the parent function is called, which is suboptimal for performance.

fb-elong:
How about the proposal below? If it is OK, you can assign the task to me.  

1.  `get_ahr`
2. `get_power`

nanxstats:
1. Perhaps `ahr_diff()` communicates better that the function is for calculating the difference between the average hazard ratio and the targeted number of events, not for getting average hazard ratio per se.
2. Maybe `get_combo_power()`? Just to be consistent with the other existing internal function `get_combo_weight()`.

nanxstats:
Fixed by #229

nanxstats:
@jdblischak do you mind helping with this when you got some time? Not urgent.

The goal is to move the inline functions `event_diff()` and `get_combo_power()` out of the upper-level function:

https://github.com/Merck/gsDesign2/blob/24fdc779a555056ecad2ef019cb455e887e51abb/R/expected_time.R#L110

https://github.com/Merck/gsDesign2/blob/4917d25086c4355d40ed455e6be21fe8d32cc8b3/R/gs_design_combo.R#L194

They are used in `uniroot()` calls.

This would drastically improve the rigorousness of these functions, because currently they used variables not defined in the function signature, thus R will search for those variables in upper level environments, which is considered dangerous. Defining them outside would enforce all invoked variables properly defined in the function signature. This would also avoid these being repeatedly redefined every time the upper-level functions are called which can be a performance hog.

PR #229, opened 2023-06-07
rename foo (#228)

elong0527:
Address suggestion in #228

### DIFF
--- a/R/expected_time.R
+++ b/R/expected_time.R
@@ -109,7 +109,7 @@ expected_time <- function(enroll_rate = tibble::tibble(
   #    build a help function    #
   # ----------------------------#
   # find the difference between  `AHR()` and different values of total_duration
-  foo <- function(x) {
+  ahr_diff <- function(x) {
     ans <- ahr(
       enroll_rate = enroll_rate, fail_rate = fail_rate,
       total_duration = x, ratio = ratio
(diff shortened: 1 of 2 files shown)


R/expected_time.R
```diff
@@ -109,7 +109,7 @@ expected_time <- function(enroll_rate = tibble::tibble(
   #    build a help function    #
   # ----------------------------#
   # find the difference between  `AHR()` and different values of total_duration
-  foo <- function(x) {
+  ahr_diff <- function(x) {
```
LittleBeannie, line 112: shall we call it `event_diff`? Since basically it calculates the difference between the targeted events and the events at a calendar time...

elong0527: sure updated.

LittleBeannie (approved):
Thanks!

nanxstats (changes requested):
I would appreciate it if you could make one small update: moving their function definitions outside of the exported functions.

fb-elong:
> I would appreciate it if you could make one small update: moving their function definitions outside of the exported functions.

Those functions are intentionally defined within a function. Because the only use case is to build a function that fit into root solving functions.  

How about we merge this PR and discuss the topic separately in #235?

nanxstats (approved):

nanxstats:
Sure! Thanks for the renaming effort and suggestions.

PR #237, opened 2023-06-18
add helper function to define expected events input

elong0527:
Enhance programming practice #238: 

Create helper functions `define_enroll_rate` to enhance user compliance. 

1. length of duration and rate must be the same. 
2. better documentation 
3. avoid user using unexpected column name in a data frame. 

If we like the idea, I will apply the same to other arguments that originally use `tibble::tibble` in future PRs.


nanxstats (approved):
I wish we had this years ago because allowing non-trivial inputs being manually constructed is a bit dangerous. So I'm going to merge this.

Issue #230, opened 2023-06-08
Remove purrr as a hard dependency


nanxstats:
purrr is currently only used once, and the particular usage has a [perfect base equivalent](https://purrr.tidyverse.org/articles/base.html#other-vector-transforms):

https://github.com/Merck/gsDesign2/blob/0f34f0145321cd1d6606c7a6f992522062ea06ea/R/expected_accural.R#L176

Therefore, I would like to see purrr being removed as a hard dependency. @elong0527 Do you mind sending a PR for this?

PR #239, opened 2023-06-18
remove purrr (#230)

elong0527:
address #230

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -40,7 +40,6 @@ Imports:
     methods,
     mvtnorm,
     npsurvSS,
-    purrr,
     rlang,
     tibble,
     tidyr,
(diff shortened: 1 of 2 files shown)


nanxstats (commented):
Would it be much simpler to write

```r
ans <- Reduce(`+`, ans)
```

than

```r
ans <- apply(do.call(cbind, ans), 1, sum)
```

?

elong0527:
> Would it be much simpler to write
> 
> ```r
> ans <- Reduce(`+`, ans)
> ```
> 
> than
> 
> ```r
> ans <- apply(do.call(cbind, ans), 1, sum)
> ```
> 
> ?

resolved.

nanxstats (approved):

PR #240, opened 2023-06-18
simplify `check_enroll_rate`

elong0527:
simplify `check_enroll_rate`


nanxstats (approved):
I changed the comment "off-label use" which has specific meanings in the medical context into "non-standard use". I also fixed a trivial spelling mistakes and improved the error messages following the tidyverse style guide.

PR #242, opened 2023-06-18
Apply `define_enroll_rate` within the R folder

elong0527:
Apply `define_enroll_rate` within the R folder


elong0527:
Please merge #240 before this PR

LittleBeannie:
May I know what's the motivation to change from `tibble` to `define_enroll_rate`? I know Keaven hesitates to API changes... considering we have a very stable API before and applied it to many studies, is it okay if we break their codes once again? 

Additionally, if we switch from `tibble` to `define_enroll_rate`, we shall also update `simtrial` for consistency.

elong0527:
User can still use `tibble` as input. I did not expect user code will break but a warning message will be provided if they keep using `tibble`. 

For a working code, the only change is from `tibble`  to `define_enroll_rate` to avoid warning message. 

Refer #237 for the benefit of using `define_enroll_rate` .

LittleBeannie:
> User can still use `tibble` as input. I did not expect user code will break but a warning message will be provided if they keep using `tibble`.
> 
> For a working code, the only change is from `tibble` to `define_enroll_rate` to avoid warning message.
> 
> Refer #237 for the benefit of using `define_enroll_rate` .

Thanks for the explanation. I installed the branch `update-enroll-rate` and run the following code with `tibble` for enroll rate, but it gives an error message...yet, if I use `define_enroll_rate`, it works.
```
gs_power_ahr(enroll_rate = tibble(
    duration = c(2, 2, 10),
    rate = c(3, 6, 9)
),
fail_rate = tibble(
    stratum = "All",
    duration = c(3, 100),
    fail_rate = log(2) / c(9, 18),
    hr = c(.9, .6),
    dropout_rate = rep(.001, 2)
),
event = c(30, 40, 50),
analysis_time = NULL,
upper = gs_b,
upar = gsDesign(
    k = 3,
    test.type = 1,
    n.I = c(30, 40, 50),
    maxn.IPlan = 50,
    sfu = sfLDOF,
    sfupar = NULL
)$upper$bound,
lower = gs_b,
lpar = c(qnorm(.1), rep(-Inf, 2)),
test_lower = TRUE,
test_upper = TRUE,
ratio = 1,
binding = FALSE,
info_scale = "h0_h1_info",
r = 18,
tol = 1e-6)
```

```
Error in UseMethod("mutate") : 
  no applicable method for 'mutate' applied to an object of class "NULL"
In addition: Warning messages:
1: In check_enroll_rate(enroll_rate) :
  Please use `define_enroll_rate()` to specify the `enroll_rate` argument.
We will enforce this requirement from the next version.
2: In check_enroll_rate(enroll_rate) :
  Please use `define_enroll_rate()` to specify the `enroll_rate` argument.
We will enforce this requirement from the next version.
3: Unknown or uninitialised column: `stratum`. 
Error in expected_time(enroll_rate = enroll_rate, fail_rate = fail_rate,  : 
  expected_time(): solution not found!
```

elong0527:
It is actually a good example to avoid `tibble`. 

You will get the same error in current CRAN version 1.0.8, because you need to define `stratum = "All"`  so the code can work.

While I updating code, I noticed some functions expected `stratum` but some are not while using `enroll_rate`. 

 If you use `define_enroll_rate`, the problem is being solved, because the default for `stratum` is "All".

PR #243, opened 2023-06-19
Define fail rate

elong0527:
add `define_fail_rate` helper function to enhance user compliance similar to #237

please merge #242 before this PR


elong0527:
Please consider review and merge this PR first. 

The `lint` pipeline is complained because of commented testing code. The problem needs to be addressed in #244.

LittleBeannie (approved):

PR #265, opened 2023-06-25
remove html files

elong0527:
remove html files from source code.


LittleBeannie:
Can we temporarily keep these files? For a long-term solution, let's save the testing code into the testthat folder.

elong0527:
Could we remove those files that can be created by Rmd files in the same folder?

LittleBeannie (approved):

PR #272, opened 2023-07-18
Update story-compute-expected-events.Rmd

LittleBeannie:


nanxstats (approved):
I only changed the fixed design vignette title to use asterisk (`*`) to represent a set of function names, consistent with Hadley's books.

PR #275, opened 2023-07-21
move old functions from inst to testthat

LittleBeannie:


LittleBeannie:
Hi @nanxstats, could you please help look at this PR? If I move the old functions (which have camel cases ) from `inst` to `testthat`, then the lintr workflow will automatically check. However, we don't want lintr to check these old functions, since we want to keep these functions as it is.

nanxstats:
@LittleBeannie Sure, let me see what I can do to pass the new lintr checks.

nanxstats (approved):
This looks great. Let's merge it.

PR #279, opened 2023-08-08
Replace expect() with expect_equal()

jdblischak:
I noticed that some tests used `expect()` when I believe they should be `expect_equal()`. The function signature of `expect()` requires the first argument to be the test, and the second is the error message to print if the test returns `FALSE`.

```R
library("testthat")
args(expect)
## function (ok, failure_message, info = NULL, srcref = NULL, trace = NULL, 
##     trace_env = caller_env()) 
## NULL

expect(1, 2)
expect(1 == 2, "they aren't equal")
## Error: they aren't equal
```

When I run the tests locally, 2 of the tests fail consistently. What is the acceptable tolerance?

```R
── Failed tests ─────────────────────────────────────────────────────────────────────────
Failure (test-independent-fixed_design.R:108:3): MaxCombo
y$analysis$power not equal to 0.9.
1/1 mismatches
[1] 0.9 - 0.9 == -2.32e-07

Failure (test-independent-fixed_design.R:140:3): RD
y$analysis$power not equal to 0.9.
1/1 mismatches
[1] 0.902 - 0.9 == 0.00163

[ FAIL 2 | WARN 2 | SKIP 0 | PASS 470 ]
```

### DIFF
--- a/tests/testthat/test-developer-expected_event.R
+++ b/tests/testthat/test-developer-expected_event.R
@@ -25,6 +25,6 @@ test_that("expected event vs gsDesign", {
     enroll_rate, fail_rate, total_duration,
     simple = TRUE
   )
-  expect(x1, x2)
-  expect(x2, x3)
+  expect_equal(x1, x2)
+  expect_equal(x2, x3)
 })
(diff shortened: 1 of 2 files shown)


nanxstats:
Good catch! For the sole purpose of making the tests pass, set:

- MaxCombo test with tolerance `testthat_tolerance() * 100`.
- RD test with tolerance `testthat_tolerance() * 2e+5`.

It might take more efforts than ideal to find out the reason for the discrepancies.

nanxstats (approved):
Awesome, thanks for the patch.

PR #281, opened 2023-08-14
Test fastlag() and use throughout package

jdblischak:
This is a follow-up from PR #280 

* Replaced `dplyr::lag()` with `fastlag()` throughout the package
* Added tests for `fastlag()`
* Added the namespace (ie `dplyr::lag()`) to one of the old functions since `lag()` is no longer in `NAMESPACE` (not currently a problem, but if we ever stopped loading dplyr in `tests/testthat.R`, it would become one)


jdblischak:
I also updated the RStudio project settings to automatically remove unnecessary whitespace. This is required by the linter, and at least in this instance, the whitespace wasn't removed when I ran styler

nanxstats (approved):
Looks good to me! 🚀

PR #285, opened 2023-08-28
278 as rtf for both gs design class and fixed design class

fukuhiro2023:


nanxstats:
I fixed the code linting issues so the lintr workflow should pass. I also added `as_rtf()` methods to `_pkgdown.yml` so the pkgdown workflow should pass.

I also fixed most of the R CMD check issues:

- Add r2rtf to Imports in `DESCRIPTION`.
- Fix check issues on `@param` and `@examples`.

The only remaining check error is from the logic so it's probably the best if you can fix it (do a pull first):

```text
Error in as_rtf.fixed_design(., title = "Fixed design", file = tempfile(fileext = ".rtf")) :
  object 'design_mtd' not found
```

Design and code style improvements:

- Rename the file path argument to `file`, make it a required argument, return the input invisibly instead of printing a message.
- Replace `|>` and `\()` to be consistent with existing code.
- rtf -> RTF.

nanxstats:
Thanks Yujie for the simple fix! I didn't notice the intention was to pipe the summary object into `as_rtf()`. Now this looks good to me.

LittleBeannie (approved):

Issue #219, opened 2023-05-03
Optimize `gs_spending_bound()`


nanxstats:
In gsDesign, the C function `hupdate` was not directly called in R, but indirectly [called](https://github.com/keaven/gsDesign/blob/7ce330dc8cf3d42d66d3c5047649c480f5fd4487/R/gsDesign.R#L139) via a C function `gsbound`, which does the memory allocation and management effectively.

In gsDesign2, the corresponding part of the [logic](https://github.com/Merck/gsDesign2/blob/6358f06466bf8f395d8c5183879ce1688d31ef26/R/gs_spending_bound.R#L212) is currently implemented in R, which could be a bottleneck to slow things down, to the extent that a prospect web interface's responsiveness will be affected.

So, at some point, we will want to rewrite `gsbound` (and `gsbound1`) in C++.

LittleBeannie:
Hi @nanxstats , is there any ideal person we can assign this ticket to? Let us do profiling first before the C++ rewrite.

nanxstats:
@LittleBeannie No one in particular came to mind, do you have a suggestion?

LittleBeannie:
I guess you are the only expert in this area in BAAMR 👍 

> @LittleBeannie No one in particular came to mind, do you have a suggestion?

elong0527:
Per team discussion, @nanxstats will drive the implementation.

LittleBeannie:
On hold for now

jdblischak:
I did some benchmarking and profiling. I don't think the R code in `gs_spending_bound()` is a major bottleneck, at least in the examples I tried. It would be helpful for me to have some extra examples of code that would typically be executed via the web interface.

First I benchmarked and profiled the example in `?gs_spending_bound`, which calls `gs_power_ahr()`. There are many bottlenecks, but `gs_spending_bound()` and `hupdate()` barely register. The biggest cumulative bottlenecks are `expected_time()`, `expected_event()`, `gs_info_ahr()`, `pw_info()`, and `ahr()` (note that these are likely nested, so I'd need to dig deeper to find the best place to optimize first)

```R
devtools::load_all()
library("microbenchmark")
library("profvis")

microbenchmark(
  gs_power_ahr = gs_power_ahr(
    analysis_time = c(12, 24, 36),
    event = c(30, 40, 50),
    binding = TRUE,
    upper = gs_spending_bound,
[...]

PR #295, opened 2023-12-14
Halve the median runtime of `gs_power_ahr()`

jdblischak:
Inspired by Issue #219, I was able to halve the runtime of the `gs_power_ahr()` example in the documentation for `gs_spending_bound()`. I achieved this by:

* converting `ahr()` to data.table
* converting almost all of `pw_info()` to data.table (there is a tricky group arrange operation at the end that I'd need to investigate more)
* updating `expected_event()` to return a data frame instead of a tibble
* updating `pw_info()` to pre-allocate lists instead of growing the data frame via `rbind()` within each loop iteration

While optimizing, I performed the following cleanups:

* added `library(gsDesign)` to `tests/testthat/test-independent-gs_power_ahr.R` so that it can be run via `devtools::test()`
* had to do some NAMESPACE rearranging to be able to use `:=` from both the data.table and rlang packages while still keeping `R CMD check` happy
* added some missing `@importFrom dplyr  filter` (I didn't get to all of them)

```R
library("microbenchmark")

remotes::install_github("Merck/gsDesign2@50067c4", INSTALL_opts = "--with-keep.source")
library("gsDesign2")
packageVersion("gsDesign2")
## [1] ‘1.1.0’

microbenchmark(
  gs_power_ahr = gs_power_ahr(
    event = c(30, 40, 50),
    binding = TRUE,
    upper = gs_spending_bound,
    upar = list(sf = gsDesign::sfLDOF, total_spend = 0.025, param = NULL, timing = NULL),
    lower = gs_spending_bound,
    lpar = list(sf = gsDesign::sfLDOF, total_spend = 0.025, param = NULL, timing = NULL)
  ),
  times = 10
)
## Unit: milliseconds
##          expr      min       lq     mean   median       uq      max neval
##  gs_power_ahr 808.6144 840.0866 940.8113 925.5892 988.4552 1201.223    10

detach("package:gsDesign2")
devtools::load_all()
packageVersion("gsDesign2")
## [1] ‘1.1.0.1’

microbenchmark(
  gs_power_ahr = gs_power_ahr(
    event = c(30, 40, 50),
    binding = TRUE,
    upper = gs_spending_bound,
    upar = list(sf = gsDesign::sfLDOF, total_spend = 0.025, param = NULL, timing = NULL),
    lower = gs_spending_bound,
    lpar = list(sf = gsDesign::sfLDOF, total_spend = 0.025, param = NULL, timing = NULL)
  ),
  times = 10
)
## Unit: milliseconds
##          expr      min       lq     mean   median       uq      max neval
##  gs_power_ahr 405.3252 427.6431 507.1905 431.4219 538.8579 779.9118    10
925 / 431
## [1] 2.146172
```


nanxstats:
This looks good from the technical perspective. We should discuss this in the meeting this week especially about using data.table.

nanxstats (approved):
Nice speed improvements. It's exciting that we have decided to use data.table to accelerate the gsDesign2 backend, too!

It seems the only explicit rlang usage (and `:=` usage) is in `R/utility_tidy_tbl.R`:

```
dplyr::rename(!!aname := !!"_alab")
```

I think it would be ideal to use an alternative approach for the line above, so that 1. `:=` is only used by data.table (ever) 2. rlang can be remove from `Imports`. 3. the `ignore_unused_imports()` workaround can be removed.

PR #299, opened 2023-12-19
Reduce interval for root finding

jdblischak:
Similar to PR #298, this PR serves as documentation of an approach we discussed. I reduced the maximum interval for the root finding by limiting it to the maximum of `analysis_time`. In the case of my benchmarking code, this reduced the max to 36. However:

* There was no detectable speed increase despite the substantial reduction in the interval
* 9 tests now fail due to minuscule differences. We could fix these failures either by relaxing the tolerance on `expect_equal()` or also updating `gs_power_ahr_()` to use the reduced interval

```R
Failure (test-developer-gs_power_ahr.R:127:3): calendar + event based cut
x1$analysis$info0 not equal to x2$info0[x2$Bound == "Lower"].
1/3 mismatches
[1] 7.5 - 7.5 == -2.37e-07

[ FAIL 9 | WARN 2 | SKIP 0 | PASS 478 ]
```

* The code to implement this isn't great. Because both `interval` and `analysis_time` are arguments to `gs_power_ahr()` and `analysis_time` can be `NULL`, the code to implement this has to be done in the argument processing. Alternatively we could change the API to set `interval  = NULL` and move this code inside the function, but again that would require more documentation

### DIFF
--- a/R/gs_power_ahr.R
+++ b/R/gs_power_ahr.R
@@ -52,7 +52,8 @@
 #'   Normally, `r` will not be changed by the user.
 #' @param tol Tolerance parameter for boundary convergence (on Z-scale).
 #' @param interval An interval that is presumed to include the time at which
-#'   expected event count is equal to targeted event.
+#'   expected event count is equal to targeted event. If `analysis_time` is
+#'   defined, the upper limit of the interval is set to its maximum.
 #'
 #' @return A tibble with columns `Analysis`, `Bound`, `Z`, `Probability`,
 #'   `theta`, `Time`, `AHR`, `Events`.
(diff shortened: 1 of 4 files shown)


R/gs_power_ahr.R
```diff
@@ -171,7 +172,7 @@ gs_power_ahr <- function(
     info_scale = c("h0_h1_info", "h0_info", "h1_info"),
     r = 18,
     tol = 1e-6,
-    interval = c(.01, 100)) {
+    interval = c(.01, if(is.null(analysis_time)) 100 else max(analysis_time))) {
```
jdblischak, line 175: Note that I used `max(analysis_time)` here simply because it is more readable than `analysis_time[length(analysis_time)]`

jdblischak:
xref: #219

PR #300, opened 2023-12-19
Use data.table for grouped sort in `pw_info()`

jdblischak:
The profiling indicated that `dplyr::group_by()` and `dplyr::arrange()` were two of the biggest bottlenecks:

image

I confirmed these were the ones in `pw_info()`:

 https://github.com/Merck/gsDesign2/blob/6d4371aef2bf205802224691e3bdff70de1ab83c/R/pw_info.R#L207-L210

In my previous PR #295, I had delayed converting this to data.table. Once I remembered that `group_by()` also orders the columns, I was able to figure it out. Here is example code to show that the two approaches are identical:

```R
set.seed(1)
d <- data.frame(
  x = sample(letters[1:3], size = 100, replace = TRUE),
  y = sample(letters[4:5], size = 100, replace = TRUE),
  z = sample(1:5, size = 100, replace = TRUE)
)

library("dplyr")
x1 <- d %>% group_by(x, y) %>% arrange(z, .by_group = TRUE) %>% ungroup()
x1 <- as.data.frame(x1)

library("data.table")
x2 <- as.data.table(d)
setorder(x2, x, y)
x2 <- x2[order(z), .SD, by = .(x, y)]
setDF(x2)

all.equal(x1, x2)
```

In repeated benchmarking of `gs_power_ahr()`, this update removes reduces the runtime by 70-140 ms

I also removed an isolated `dplyr::transmute()` from `gs_info_rd()` and updated the documentation to reflect that now a data frame is returned


jdblischak:
xref: #219

keaven (approved):
Thanks for the update

PR #305, opened 2024-01-03
Enhance documention in `define.R`

### LINKED ISSUE #302: Improve documentation for `define_enroll_rate()`
Suggestions from Keaven:

> Before we had a prevalence for each stratum SEPARATE from enrollment rates. This assumed there would be a set of overall piecewise enrollment rates and then a multinomial probability would be used to derive stratum randomly for each observation. The help file does not document behavior well.
>
> In particular, it is not clear what the 2nd example in the help file for `define_enroll_rate()` even means.

(I wonder if `define_fail_rate()` also needs similar attention)

elong0527:
Fixes #302 

- provide function description.
- update example in `define_fail_rate`.

### DIFF
--- a/R/define.R
+++ b/R/define.R
@@ -18,12 +18,16 @@
 
 #' Define enrollment rate
 #'
-#' Define the enrollment rate of subjects for a study as following a piecewise exponential distribution.
-#' The `duration` are ordered piecewise for a duration equal to \eqn{t_i - t_{i-1}}.
-#' where \eqn{0=t_0<t_i\ldots,t_M=\infty}.
+#' Define the enrollment rate of subjects for a study as following a piecewise
+#' exponential distribution.
+#'
+#' @details
+#' The `duration` are ordered piecewise for a duration equal to
+#' \eqn{t_i - t_{i-1}}, where \eqn{0 = t_0 < t_i < \cdots < t_M = \infty}.
 #' The enrollment rates are defined in each duration with the same length.
 #'
-#' For a study with multiple strata, different duration and rate can be specified in each stratum.
+#' For a study with multiple strata, different duration and rate can be
+#' specified in each stratum.
 #'
 #' @param duration A numeric vector of ordered piecewise study duration interval.
 #' @param rate A numeric vector of enrollment rate in each `duration`.
@@ -46,7 +50,6 @@
 #'   rate = c((1:4) / 3, (1:4) / 2, (1:4) / 6),
 #'   stratum = c(array("High", 4), array("Moderate", 4), array("Low", 4))
 #' )
-#'
 define_enroll_rate <- function(
     duration,
     rate,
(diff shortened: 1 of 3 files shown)


R/define.R
```diff
@@ -18,6 +18,10 @@
 
 #' Define enrollment rate
 #'
+#' The function define subject enrollment rate for a study.
```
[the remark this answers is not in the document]
elong0527: The current define and AHR functions did not consider different prevalence in each stratum. From the source code of the AHR function. It seems only work for equal weight. Feel a different issue is required to address unequal prevalence in each stratum before we can document the details 

The 2nd example has been updated, I feel the example is self explained.

LittleBeannie:
Here is an example of stratified design: https://merck.github.io/gsDesign2/articles/story-ahr-under-nph.html#design-scenario-1.

R/define.R
nanxstats, line 21: 1. define -> define**s**
2. Avoid starting the sentence with obvious stuff like "This function..." or "This package...".

R/define.R
```diff
@@ -18,6 +18,10 @@
...
+#' It also supports stratified design that have different enrollment rate in
```
nanxstats, line 22: stratified design**s** that have different enrollment rate**s**

R/define.R
```diff
@@ -74,20 +78,25 @@ define_enroll_rate <- function(
   df
 }
 
-#' Define fail rate
+#' Define failure rate
+#'
+#' The function define subject failure rate for a study with two treatment groups.
```
nanxstats, line 83: 1. define**s**
2. "This function"

R/define.R
```diff
@@ -74,20 +78,25 @@ define_enroll_rate <- function(
...
+#' It also supports stratified design that have different failure rate in
```
nanxstats, line 84: stratified design**s** that have different enrollment rate**s**

R/define.R
```diff
@@ -74,20 +78,25 @@ define_enroll_rate <- function(
...
+#' each stratum.
+#'
 #'
 #' @param duration A numeric vector of piecewise study duration interval.
-#' @param fail_rate A numeric vector of failure rate in each `duration`.
+#' @param fail_rate A numeric vector of failure rate in each `duration` in the control group.
 #' @param dropout_rate A numeric vector of dropout rate in each `duration`.
-#' @param hr A numeric vector of hazard ratio.
+#' @param hr A numeric vector of hazard ratio between case and control group.
```
nanxstats, line 91: gsDesign and gsDesign2 both used "treatment group" instead of "case group" consistently. There was not a single "case group" usage.

R/define.R
```diff
@@ -97,11 +106,11 @@ define_enroll_rate <- function(
 #'
 #' # Define enroll rate with stratum
```
nanxstats, line 107: For both examples, please please consider explaining in plain language what are being defined - this is because something looks natural to you doesn't mean it's "self-explanatory" to all - maybe it's because you are more familiar with the spec. 🙏

nanxstats (changes requested):
I made a few comments. Please consider patching them up. I agree with @LittleBeannie that some aspects of the spec deserve better articulation.

R/define.R
elong0527: The doc are updated. Please take another look to see if it is clear.

LittleBeannie (approved):
Thanks for the updates, Yilong! I just approved this PR and will merge it after Nan's approval.

nanxstats (approved):
I appreciate the updates - I also improved a few details. Let's merge this.

PR #306, opened 2024-01-08
Add reference sections "Piecewise exponential" and "Expected ..."

### LINKED ISSUE #258: re-org the reference page in the pkgdown website
- [ ] Start a new section of "piecewise exponential" and save `ppwe` and `s2pwe` there
- [ ] Start a new section of "expected ..." and save `expected_event`, `expected_time` and `expected_accrual` there

jdblischak:
Closes #258 

* Please feel free to suggest more detailed descriptions for these sections
* I removed the functions from low-level helpers when I moved them to their dedicated section. If you'd prefer to also leave them there, I can add them back. Duplication isn't a problem here since we want users to be able to find the functions in the section they expect

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -99,15 +99,25 @@ reference:
   contents:
     - gs_b
     - gs_spending_bound
+- title: "Expected ..."
+  desc: >
+    Functions for computing trial events.
+  contents:
+    - expected_event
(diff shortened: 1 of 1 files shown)


LittleBeannie (approved):
Thank you for cleaning it up!

PR #307, opened 2024-01-08
Remove {rlang} to simplify NAMESPACE

jdblischak:
I introduced {data.table} in PR #295 to increase the computational efficiency. Both {data.table} and {rlang} define the operator `:=`, which complicates the package NAMESPACE. Motivated by @nanxstats comment in https://github.com/Merck/gsDesign2/pull/295#pullrequestreview-1785047546, I was able to simplify the NAMESPACE by removing the few uses of {rlang}

https://github.com/Merck/gsDesign2/blob/40a4c8b1ee64adf0693edd3f361ba5e928de0db0/R/utility_tidy_tbl.R#L81-L83

I didn't need to add any new tests because the output table returned by this function is already well-tested:

https://github.com/Merck/gsDesign2/blob/40a4c8b1ee64adf0693edd3f361ba5e928de0db0/tests/testthat/test-independent-utility_tbl.R#L17-L34

Also note that this function is not exported, and its documentation file is suppressed with `@noRd`

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: gsDesign2
 Title: Group Sequential Design with Non-Constant Effect
-Version: 1.1.0.2
+Version: 1.1.0.3
 Authors@R: c(
     person("Keaven", "Anderson", email = "keaven_anderson@merck.com", role = c("aut")),
     person("Yilong", "Zhang", email = "elong0527@gmail.com", role = c("aut")),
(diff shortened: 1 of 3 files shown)


nanxstats (approved):
Thanks so much for this update.

I'm so happy to see the single update made the dependency graph lighter, and I think the updated logic is much easier to read - I often feel that I start to lose my mind after reading a dozen `!!`, `!!!`, and `:=` operators... 😂

jdblischak:
> I often feel that I start to lose my mind after reading a dozen !!, !!!, and := operators...

I feel the same! I was actually surprised with how simple the replacement code was. I thought all the `!!` were doing something more complex

PR #311, opened 2024-01-12
Update copyright year

nanxstats:
This PR updates the copyright year in the code header and pkgdown site footer to 2024.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -632,7 +632,7 @@ state the exclusion of warranty; and each file should have at least
 the "copyright" line and a pointer to where the full notice is found.
 
     gsDesign2: Group Sequential Design with Non-Constant Effect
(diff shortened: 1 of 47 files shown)


jdblischak:
I'm neutral on this change. I've seen arguments that you need to use a date range, which is what I typically do for my software, but then again GitHub only has 2024 listed in its footer, so it's hard to argue with that precedent

LittleBeannie (approved):

nanxstats:
> I'm neutral on this change. I've seen arguments that you need to use a date range, which is what I typically do for my software, but then again GitHub only has 2024 listed in its footer, so it's hard to argue with that precedent

That makes sense in some cases, too.

Simple reason here: we were asked to strictly follow this (most recently updated year) format so no need to think...

Issue #235, opened 2023-06-14
Location of functions for root solving [lower priority] 


fb-elong:
Place to discuss comments in #229

>I would appreciate it if you could make one small update: moving their function definitions outside of the exported functions.

nanxstats:
Following the discussion today: the reason that the inner function is harder to move outside the outer function: the inner function is using variables outside of it's scope but within the outer function's scope. To move it out, additional arguments need to be added.

elong0527:
Removed my assignment as no short-term plan to complete it.

nanxstats:
@jdblischak do you mind helping with this when you got some time? Not urgent.

The goal is to move the inline functions `event_diff()` and `get_combo_power()` out of the upper-level function:

https://github.com/Merck/gsDesign2/blob/24fdc779a555056ecad2ef019cb455e887e51abb/R/expected_time.R#L110

https://github.com/Merck/gsDesign2/blob/4917d25086c4355d40ed455e6be21fe8d32cc8b3/R/gs_design_combo.R#L194

They are used in `uniroot()` calls.

This would drastically improve the rigorousness of these functions, because currently they used variables not defined in the function signature, thus R will search for those variables in upper level environments, which is considered dangerous. Defining them outside would enforce all invoked variables properly defined in the function signature. This would also avoid these being repeatedly redefined every time the upper-level functions are called which can be a performance hog.

jdblischak:
`event_diff()` was straightforward, so I started with it. See #312

I'm a bit more cautious about `get_combo_power()` because the call to `uniroot()` in `gs_design_combo()` is more complex. It allows passing arguments via `...` in the top-level function signature all the way to `mvtnorm::pmvnorm()` via the `uniroot()` call

https://github.com/Merck/gsDesign2/blob/59e8a0315394a32d7437cab503da14b536fe1e1b/R/gs_design_combo.R#L26
https://github.com/Merck/gsDesign2/blob/59e8a0315394a32d7437cab503da14b536fe1e1b/R/gs_design_combo.R#L225

However, as a quick test, I deleted the `...` from the `uniroot()` call and ran the tests. Everything passed, which means this feature isn't being tested, and thus we wouldn't know if my change broke this behavior. My plan is to write a test first. If anyone knows of a realistic use case for passing an argument to `mvtnorm::pmvnorm()`, I'd appreciate any suggestions

nanxstats:
[...]

PR #312, opened 2024-01-16
Define event_diff() as a separate internal function

jdblischak:
Motivation: https://github.com/Merck/gsDesign2/issues/235#issuecomment-1884321766


R/expected_time.R
```diff
@@ -120,3 +112,24 @@ expected_time <- function(
     return(ans)
   }
 }
+
+#' Find the difference between `ahr()` and different values of `total_duration`
```
LittleBeannie, line 116: I personally suggest rewording it as follows: 'Considering the enrollment rate, failure rate, and randomization ratio, calculate the difference between the targeted number of events and the accumulated events at time `x`.' Please review and let me know if you agree with this rephrased version.

LittleBeannie (commented):
Hi @jdblischak , thanks for the updates! I only have 1 minor comment below.

R/expected_time.R
jdblischak: I agree that's a great title! I just pushed the change

LittleBeannie (approved):

nanxstats (approved):
Awesome! I appreciate this refactor that makes `expected_time()` more robust.

PR #314, opened 2024-01-16
Upgrade to roxygen2 7.3.0

### LINKED ISSUE #313: roxygen2 7.3.0 breaks NAMESPACE
While working on #312, I noticed that the latest roxygen2 was deleting `:=` from {data.table} and `%>%` from {dplyr} from `NAMESPACE`.

```R
==> devtools::document(roclets = c('rd', 'collate', 'namespace'))

ℹ Updating gsDesign2 documentation
✖ pw_info.R:75: @importFrom Excluding unknown export in from data.table:
  `":="`.
✖ pwe.R:153: @importFrom Excluding unknown export in from dplyr: `"%>%"`.
ℹ Loading gsDesign2
✖ pw_info.R:75: @importFrom Excluding unknown export in from data.table:
  `":="`.
✖ pwe.R:153: @importFrom Excluding unknown export in from dplyr: `"%>%"`.
Documentation completed
```

roxygen2 7.3.0 was recently [released](https://www.tidyverse.org/blog/2024/01/roxygen2-7-3-0/) on 2024-01-11. Fortunately another developer has already reported the problem in https://github.com/r-lib/roxygen2/issues/1570. While we wait on a fix, please build the documentation with roxygen2 7.2.3

jdblischak:
Closes #313 

It turns out that if I remove the quotes around `:=` and `%>%`, then the `NAMESPACE` is correctly generated by both roxygen2 7.2.3 and 7.3.0

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -60,4 +60,4 @@ VignetteBuilder:
 LinkingTo:
     Rcpp
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.2.3
+RoxygenNote: 7.3.0
--- a/R/pw_info.R
+++ b/R/pw_info.R
@@ -72,7 +72,7 @@
 #' }
 #' \if{html}{The contents of this section are shown in PDF user manual only.}
 #'
-#' @importFrom data.table ":=" as.data.table copy first last rbindlist setDF
+#' @importFrom data.table := as.data.table copy first last rbindlist setDF
 #'                        setorderv
 #'
 #' @export
--- a/R/pwe.R
+++ b/R/pwe.R
@@ -150,7 +150,7 @@ ppwe <- function(x, duration, rate, lower_tail = FALSE) {
 #'  }
 #' \if{html}{The contents of this section are shown in PDF user manual only.}
 #'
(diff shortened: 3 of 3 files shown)


nanxstats (approved):
Although I'm often reluctant to suggest changes that would lead to [large-scale changes](https://abseil.io/resources/swe-book/html/ch22.html) (such as requiring 10+ PRs to update all packages), looks like we may need a path forward here before the next roxygen2 release, and this might be the only way.

I hope roxygen2 > 7.3.0 will fix the bug to support with quotes and keep supporting without quotes so we don't need to change this back. 🤞

jdblischak:
> requiring 10+ PRs to update all packages

I view this as a temporary problem. Thus we only need to update packages that are being actively developed

> I hope roxygen2 > 7.3.0 will fix the bug to support with quotes and keep supporting without quotes so we don't need to change this back.

That's why I tested with roxygen2 7.2.3. The unquoted version worked before, currently works, and will continue to work. And I'm less worried about future problems since the PR https://github.com/r-lib/roxygen2/pull/1574 is adding tests for all the various use cases (no quotes, single quotes, double quotes, and back ticks)

PR #315, opened 2024-01-17
Pass args via `...` to `mvtnorm::pmvnorm()`

jdblischak:
To aid our discussion this afternoon. If we want to enable passing parameters to `mvtnorm::pmvnorm()`, this PR fixes that behavior. If we decide to keep this functionality, it would be nice to add some tests of realistic use cases so that this can be caught in the future.

xref: https://github.com/Merck/gsDesign2/issues/235#issuecomment-1896298750

Note that the call to `pmvnorm_combo()`  in `gs_bound()` does pass the `...`:

https://github.com/Merck/gsDesign2/blob/f4a9e8dc03e6383067912262d447c50e7d060d1a/R/utility_combo.R#L481-L486

What this PR does is add the `...` to the 2 calls in `gs_prob_combo()`

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: gsDesign2
 Title: Group Sequential Design with Non-Constant Effect
-Version: 1.1.0.4
+Version: 1.1.0.5
 Authors@R: c(
     person("Keaven", "Anderson", email = "keaven_anderson@merck.com", role = c("aut")),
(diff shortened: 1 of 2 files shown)


jdblischak:
Note that the CI failures are due to a problem with installing the dependency {withr}

nanxstats (approved):
I approve this approach. Looking forward to those new tests! ✨

PR #317, opened 2024-01-18
Test that args are passed via `...` to `mvtnorm::pmvnorm()`

jdblischak:
I decided to use the argument `seed`. I confirmed that if I interrupt the chain of `...` by deleting them from one of the intermediate function calls, then these tests fail.

xref: Followup to #315, motivation in https://github.com/Merck/gsDesign2/issues/235#issuecomment-1897782719


nanxstats (approved):
Ha! I like the simple solution. I didn't know this `seed` argument existed before.

Looking at the news file, it turns out `seed` was first added in mvtnorm 1.2-0 released on 2023-06-02.

PR #318, opened 2024-01-18
Define get_combo_power() as a separate internal function

### LINKED ISSUE #235: Location of functions for root solving [lower priority]

jdblischak:
Closes #235

xref: #312, #315, #317

### DIFF
--- a/R/gs_design_combo.R
+++ b/R/gs_design_combo.R
@@ -186,21 +186,6 @@ gs_design_combo <- function(
     min_info_frac <- apply(do.call(rbind, info_frac), 2, min)
   }
 
-  # Function to calculate power
-  get_combo_power <- function(n, beta, ...) {
-    # Probability Cross Boundary
-    prob <- gs_prob_combo(
-      upper_bound = bound$upper,
-      lower_bound = bound$lower,
-      analysis = info_fh$analysis,
-      theta = theta_fh * sqrt(n),
-      corr = corr_fh,
-      algorithm = algorithm, ...
-    )
-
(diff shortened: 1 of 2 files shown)


nanxstats (approved):
I don't think I see any obvious issues, so let's merge this.

I like in the refined `uniroot()` call how you correctly placed `...` first to be passed to `pmvnorm()` and then `extendInt = "yes"` as further arguments to `uniroot()`. 👍

R/gs_design_combo.R
```diff
@@ -341,3 +339,27 @@ gs_design_combo <- function(
 
   return(output)
 }
+
+#' Function to calculate power
```
LittleBeannie, line 343: Could we include some additional description for this help function? While I understand that it is an internal function and an immediate documentation PR may not be necessary, it would still be beneficial to have it.

"This function calculates the difference between the derived power and the targeted power (1 - beta), based on the provided sample size, upper and lower boundaries, and treatment effect."

LittleBeannie (commented):
Hello @jdblischak, I completely understand that this PR has already been merged. However, I do have a very minor comment regarding the documentation that I wanted to bring to your attention.

R/gs_design_combo.R
jdblischak: @LittleBeannie thanks! Please see https://github.com/Merck/gsDesign2/pull/320/commits/f81a832e94a0f8bcfae3962761a38bc8e22b4c44

PR #325, opened 2024-01-25
Add {stats} to Imports

jdblischak:
I noticed that {stats} is in `NAMESPACE` but not `DESCRIPTION`

https://github.com/Merck/gsDesign2/blob/187edf339dcb6c953aa9dd3998560310803c1cf1/NAMESPACE#L80-L83


nanxstats (approved):
Nice finding - always good to be consistent.

PR #326, opened 2024-01-25
Enable manual trigger of GitHub Actions workflows

jdblischak:
Some minor updates to our CI files

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -5,6 +5,7 @@ on:
     branches: [main, master]
   pull_request:
     branches: [main, master]
+  workflow_dispatch:
 
 name: R-CMD-check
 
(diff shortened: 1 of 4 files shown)


nanxstats (commented):
I'm a little worried that this is deviating too far from the standard GHA workflow in r-lib/actions and it's generally challenging to maintain nonstandard stuff.

I propose to use a simple solution instead: we restore the standard GHA workflow from upstream, add you as a collaborator of the repo, and you can create development branches here directly. What do you think?

nanxstats (approved):
Thanks for making the workflows have the right level of permissions!

jdblischak:
Another quick follow-up: confirmed that the pkgdown site was still pushed to gh-pages in https://github.com/Merck/gsDesign2/commit/bd3430ecc323884b6eaf3de0b73659c9fbbb0dab

PR #327, opened 2024-01-26
Restore operator quotations and update to roxygen2 7.3.1

nanxstats:
This PR reverts the changes in https://github.com/Merck/gsDesign2/pull/314 by adding the quotes back as the parser bug in roxygen2 7.3.0 has been fixed in roxygen2 7.3.1 (on CRAN 2024-01-22), and reran roxygen2 to update the `RoxygenNote` field in `DESCIRPTION` to use 7.3.1.

Changing this back ensures the convention to import special operators like `:=` or `%>%` is consistent with both

1. The generated `NAMESPACE` file in this package.
2. All other packages that have been quoting them all along.

so I think this is worth it.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -61,4 +61,4 @@ VignetteBuilder:
 LinkingTo:
     Rcpp
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.3.0
+RoxygenNote: 7.3.1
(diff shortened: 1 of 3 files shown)


nanxstats:
Note that although running this using roxygen2 7.3.0 will result in those parsing errors, this message will be shown at the top:

```
ℹ Updating gsDesign2 documentation
✖ Installed roxygen2 is older than the
  version used with this package
ℹ You have "7.3.0" but you need "7.3.1"
```

which tells people what to do - upgrade.

jdblischak (approved):

PR #328, opened 2024-01-31
add error message checking the integer sample size

LittleBeannie:


nanxstats:
Nice... Maybe `is.wholenumber()` in gsDesign is a useful one-liner internal function to copy here for repeated use:

https://github.com/keaven/gsDesign/blob/bd70f66e346603229b46de2a7a84f0a714857ba1/R/toInteger.R#L159

LittleBeannie:
> Nice... Maybe `is.wholenumber()` in gsDesign is a useful one-liner internal function to copy here for repeated use:
> 
> https://github.com/keaven/gsDesign/blob/bd70f66e346603229b46de2a7a84f0a714857ba1/R/toInteger.R#L159

Thanks for the suggestion. Yeah, gsDesign2 exactly apply the same logic as in `gsDesign:::is.wholenumber`. Since it is a short function, and we only use it once, shall we just use it within the if condition?

nanxstats (approved):
I ported and added it as `is_wholenumber()` to encourage code reuse.

PR #329, opened 2024-02-02
289 vignette of the information under the null and alternative hypothesis

LittleBeannie:


nanxstats (approved):
This looks super clear. Thanks for adding the vignette!

PR #332, opened 2024-02-08
Qualify namespaces in tests to avoid `library()` calls

nanxstats:
Following the recommendation in [R Packages](https://r-pkgs.org/testing-basics.html#test-organisation) (also tidyverse and tidymodels conventions), this PR removes the extra `library()` calls from `tests/testthat.R` and individual tests, and qualify the namespaces explicitly.

The involved namespaces and functions include (might not be exhaustive):

```
- gsDesign: gsDesign gsSurv gsBoundSummary sfLDOF
- tibble: tibble
- dplyr: filter mutate transmute full_join group_by summarize ungroup
  first last lag lead select arrange desc right_join rename
```

### DIFF
--- a/tests/testthat.R
+++ b/tests/testthat.R
@@ -1,6 +1,4 @@
 library(testthat)
-library(gsDesign)
 library(gsDesign2)
-library(dplyr)
 
 test_check("gsDesign2")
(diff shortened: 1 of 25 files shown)


LittleBeannie (approved):
Thanks, @nanxstats !

PR #335, opened 2024-02-13
331 efficacy and futility boundary update

### LINKED ISSUE #331: Efficacy/Futility boundary update
Add vignette on efficacy/futility boundary update.

LittleBeannie:
Closes #331


LittleBeannie:
@nanxstats , please review and edit if you find anywhere inappropriate, but let's wait for Keaven's final approval for merge.

nanxstats (commented):
This looks nice. Besides the style improvements, I did two things worth mentioning:

I only kept the `library(gsDesign2)` call and removed all other `library()` calls, especially for gsDesign and gt, to avoid the function name masking situations on `as_gt()` and `as_rtf()`, and possibly others. Then I qualified the namespaces with `::` in the code chunks explicitly. This could help minimize potential future issues rooted from "loading packages and change the R landscape", especially when serving as a robust template.

Also, since most code chunks use `echo = FALSE`, I set that option in the beginning globally and use `echo = TRUE` in the chunks that do need to echo.

nanxstats (approved):
This looks good to me. Thanks!

PR #336, opened 2024-02-13
Update vignette for spending bounds under nph

LittleBeannie:
This PR:

- Updated the vignette "trial design with spending under NPH".
- Extended the vignette "computing spending boundaries in group sequential design" to compare derivation of different spending bounds using gsDesign2 and gsDesign.

### DIFF
--- a/.lintr
+++ b/.lintr
@@ -11,6 +11,9 @@ exclusions:
     list(
       "inst/",
       "tests/testthat/old_function/",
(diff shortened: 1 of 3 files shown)


nanxstats (approved):
Thanks for the **substantial** improvement!

I excluded the linting rule for the vignette with camel case variable names, as it was intended used for comparison and contrast purposes.

PR #338, opened 2024-02-15
Run `as_gt()` examples conditionally

### LINKED ISSUE #334: Reasonable `as_gt()` code example output
Currently, the pkgdown page for [`as_gt()`](https://merck.github.io/gsDesign2/reference/as_gt.html) outputs 7,000 lines of raw HTML.

It looks like gt is doing something special for the code examples section. So perhaps check how gtsummary handles this.

nanxstats:
Fixes #334 

I used `@examplesIf` to only run the `as_gt()` examples when it is either in interactive mode OR not rendered by pkgdown.

```r
#' @examplesIf interactive() && !identical(Sys.getenv("IN_PKGDOWN"), "true")
```

This will make the examples section in the pkgdown site not showing the 7,000 lines of raw HTML.

---

I did not choose to capture the tables as images and create an "example output" section to link them (what gtsummary does) because:

- gt outputs is not the main theme of this package so it's less important to show what the output looks like.
- Since we have 11 tables here, the saved PNG files will bloat up the bundled package size by at least 1 Mb.
- One needs to regenerate the images using a script if the examples are updated. As people tend to forget about things outside of the standard workflow, it could become easily out of sync or a maintenance burden.

### DIFF
--- a/R/as_gt.R
+++ b/R/as_gt.R
@@ -32,9 +32,8 @@ as_gt <- function(x, ...) {
 #'
 #' @export
 #'
-#' @examples
+#' @examplesIf interactive() && !identical(Sys.getenv("IN_PKGDOWN"), "true")
 #' library(dplyr)
-#' library(tibble)
 #'
 #' # Enrollment rate
 #' enroll_rate <- define_enroll_rate(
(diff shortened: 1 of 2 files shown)


LittleBeannie (approved):
Thanks, @nanxstats !

PR #345, opened 2024-02-22
add `creat_arm` by lose `sum(accr_para) == 1` to `is_wholenumber(sum(accr_para) - 1)`

LittleBeannie:


nanxstats (commented):
I appreciate the idea but the best way is to contribute a patch to the upstream dependency.

Copying an entire function (implementation) with one thing changed instead of coding against the interface is problematic. This is because if the upstream dependency changes their internal implementations, those changes can easily break you package, as your modified copy is not updated.

The approach here will likely become technical debt and add extra burden to refactor.

nanxstats:
Your code also loosened the original condition. The new logic only checked if the number is almost any integer:

```r
if (length(accr_interval) > 2 && !is_wholenumber(sum(accr_param) - 1))
```

but not checking if the number is almost `1L` specifically. It should be something like this:

```r
is_almost_k <- function(x, k, tol = .Machine$double.eps^0.5) abs(x - k) < tol

if (length(accr_interval) > 2 && !is_almost_k(sum(accr_param), k = 1L))
```

nanxstats (approved):
Following today's discussion, let's merge this so we have a working version for the short term, and create an issue #346 on reverting it back once npsurvSS is patched in the future.

PR #347, opened 2024-02-25
Make tests self-contained

nanxstats:
As recommended by [R Packages](https://r-pkgs.org/testing-design.html), this PR makes all existing tests self-contained by creating helper functions to wrap the top-level logic that used to have awkward "test file scope".

Now all test files only contain individual `test_that()` blocks. There are a few exceptions where a few test files containing one/two-liner batch `source()` calls or running computationally expensive helper functions at the top.

### DIFF
--- a/.lintr
+++ b/.lintr
@@ -21,7 +21,7 @@ exclusions:
       "tests/testthat/helper.R" = list(
         object_name_linter = Inf
       ),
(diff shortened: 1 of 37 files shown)


LittleBeannie (approved):
Thanks, @nanxstats !

jdblischak (approved):

PR #348, opened 2024-02-26
Improve readme code example

nanxstats:
This PR improves the code example in `README.Rmd`:

- Update code so it works with under the latest API (`fixed_design()` -> `fix_design_ahr()`).
- Remove unnecessary `library()` calls and only keep gsDesign2. Qualify the other namespaces explicitly.
- Add code chunks for table rendering purposes and remove `gt::as_raw_html()` from the code chunk for display to reduce confusion and visual clutter.
- Render and refresh `README.md` so the output reflects the latest table output format.

Also, importantly:

- Use the native pipe `|>` (as in the purrr readme) to avoid the magrittr dependency where appropriate and possible. This means readme can use `|>`, but not roxygen2 code examples, which would trigger `Depends: R >= 4.1.0` checks.
- In the relevant code chunks, `eval` is set to `getRversion() >= "4.1"` so the code chunks will only run when `|>` exists.

### DIFF
--- a/README.Rmd
+++ b/README.Rmd
@@ -71,9 +71,6 @@ enrollment duration of 12 months with exponential inter-arrival times.
 
 ```{r, message=FALSE, warning=FALSE}
 library(gsDesign2)
-library(gsDesign)
-library(dplyr)
-library(gt)
 
 # Basic example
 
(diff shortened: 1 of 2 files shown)


LittleBeannie (approved):
Thanks for the edits!

PR #353, opened 2024-03-01
Add vignette link in readme

nanxstats:
This PR adds the gt vignette link to readme so people can click through it directly instead of looking for it.


LittleBeannie (approved):
Thanks, @nanxstats !

PR #354, opened 2024-03-01
Make Markdown output fully reproducible by removing gt div

nanxstats:
This PR further post-processes the gt outputs in `README.md` to remove the associated `<div>` blocks (previously with random id) to make the output **fully** reproducible.

I also verified that the rendered readme will still have the same look on both GitHub and pkgdown.


LittleBeannie (approved):
Thanks, @nanxstats !

PR #355, opened 2024-03-01
Move reference implementation as test helper files

nanxstats:
Following the recommendations in the [designing your test suite](https://r-pkgs.org/testing-design.html) chapter in the R Packages book, this PR moves the gsDesign2 reference implementation used in tests, which was previously stored as `tests/testthat/old_function/*.R`, to `tests/testthat/helper-reference-*.R`.

This makes them directly available to all tests and avoids using the `source()` function "that should almost never appear below `tests/testhat/`" in test files.


LittleBeannie (approved):
Thanks, @nanxstats

PR #369, opened 2024-04-01
368 bug of to integer when the study duration is 100

LittleBeannie:


nanxstats (approved):
I updated the gt testing snapshots and now those snapshot tests should pass. This might be caused by a gt LaTeX renderer change from upstream.

PR #337, opened 2024-02-14
Add independent tests for `as_gt()`

DMuriuki:
Initial PR

PR #371, opened 2024-04-02
Skip running gt snapshot tests on CRAN

nanxstats:
This PR enhances https://github.com/Merck/gsDesign2/pull/337 by adding a `skip_on_cran()` condition to the gt snapshot tests.

The reason behind this is that once new versions of gt is updated on CRAN, it might render slightly different outputs, just like what happened in https://github.com/Merck/gsDesign2/pull/369. This would make these tests prone to failure on CRAN for reasons outside of our control. Therefore, we are better off skipping them on CRAN.

They will still run in GitHub Actions so we are aware whether they work or not.

### DIFF
--- a/tests/testthat/test-independent_as_gt.R
+++ b/tests/testthat/test-independent_as_gt.R
@@ -30,6 +30,8 @@ test_that("fail_rate produces the expected output", {
 })
 
 test_that("Snapshot test for fixed_design summary as_gt", {
+  skip_on_cran()
+
   enroll_rate <- define_enroll_rate(duration = 18, rate = 20)
   fail_rate <- define_fail_rate(
     duration = c(4, 100),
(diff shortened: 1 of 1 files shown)


LittleBeannie (approved):
Thanks!

nanxstats:
@yihui I think we can agree on the problem (upstream dependencies make breaking changes that will break downstream packages tests easily), but I respectfully disagree with the recommendation on not using `skip_on_cran()`.

The rationale has been explained in #371, but maybe it can use more clarity: having to make frequent CRAN updates due to upstream dependency changes adds unnecessary time and efforts burden to maintainers (see previous examples: https://github.com/Merck/simtrial/pull/234, https://github.com/keaven/gsDesign/pull/115). Although we welcome gt authors to send PRs to fix the problems, it is **us** who still have to make CRAN releases, and you know that is not as trivial as releasing a Python package to PyPI due to the high standards. We also don't want to be a potential blocker for gt's CRAN releases.

PR #378, opened 2024-04-10
377 gs update ahr change of alpha at the design stage when multiplicity graph changes

### LINKED ISSUE #377: `gs_update_ahr`: change of alpha at the design stage when multiplicity graph changes
In clinical trial design, multiple hypotheses share a total alpha (say, 0.025). A single hypothesis may have several alpha. Specifically, it starts with an initial alpha. If the multiplicity graph changes due to some not-rejected hypotheses, its alpha may change. 

The `gs_update_ahr` should incorporate this scenario.

LittleBeannie:
Closes #377


nanxstats:
To to pass the tests, use the latest gt version to generate the snapshots. Your local gt version is too old - that's where the discrepancy of `\addlinespace[2.5pt]` in the snaps came from.

nanxstats (approved):

Issue #382, opened 2024-04-26
`to_integer()` output incorrect timing under calendar-based spending


LittleBeannie:
The following codes generate a calendar-based spending design. The output directly from `gs_design_ahr()` provides a correct timing of 60%. However, the output after the `to_integer()` gives the wrong timing.

```
library(gsDesign)
library(gsDesign2)

strata <- c("A", "B")   # Population strata
prevalence <- c(.3, .7)     # Prevalence of each stratum
hr <- c(.7, .8)             # Hazard ratio within each stratum
control_median <- c(8, 10) # Control group median in months by stratum
enrollment_duration <- 12   # Expected enrollment duration 
trial_duration <- 30        # Targeted trial duration
analysis_time <- c(18, 30)  # Targeted time of analyses
dropout_rate <- c(0.001, 0.001)

x <- gs_design_ahr(
  enroll_rate = define_enroll_rate(duration = rep(enrollment_duration, 2), 
                                   rate = prevalence, stratum = strata),
  fail_rate = define_fail_rate(
    duration = rep(Inf, 2), 
    fail_rate = log(2)/control_median, 
    hr = hr, 
    dropout_rate = dropout_rate,
    stratum = strata),
  alpha = 0.025,
  beta = 0.1,
  info_frac = NULL,
  analysis_time = analysis_time,
  ratio = 1,
  binding = FALSE,
  upper = gs_spending_bound,
  upar = list(sf = gsDesign::sfLDOF, total_spend = 0.025, param = NULL, 
              timing = analysis_time / last(analysis_time)),
  lower = gs_b,
  lpar = c(-Inf, -Inf, -Inf),
  h1_spending = TRUE,
  test_upper = TRUE,
  test_lower = FALSE,
  info_scale = "h0_info"
) 

x$input$upar$timing
(x |> to_integer())$input$upar$timing
```

PR #383, opened 2024-04-26
Correct timing when it is calendar-based spending

LittleBeannie:


jdblischak (changes requested):
Since you are changing the behavior of `to_integer()`, could you please bump the build number to 1.1.2.3?

https://github.com/Merck/gsDesign2/blob/eb62b764b48f1e9951751db1a6b630d700d78ff8/DESCRIPTION#L3

Also, I recommend converting the example in https://github.com/Merck/gsDesign2/issues/382 to a test so that we can be confident that the error remains fixed in the future.

LittleBeannie:
> Since you are changing the behavior of `to_integer()`, could you please bump the build number to 1.1.2.3?
> 
> https://github.com/Merck/gsDesign2/blob/eb62b764b48f1e9951751db1a6b630d700d78ff8/DESCRIPTION#L3
> 
> Also, I recommend converting the example in #382 to a test so that we can be confident that the error remains fixed in the future.

Good suggestions, agreed! I will get your comments addressed in another commit.

jdblischak (approved):

nanxstats (approved):
(I simply ran `styler:::style_active_file()` on the change files to fix the code style problems.)

PR #386, opened 2024-05-01
385 delete lintr workflow

LittleBeannie:


nanxstats (approved):
(I added `.lintr` back because it's still useful for running the linter locally.)

For context, @LittleBeannie and I had a discussion and realized that due to how this code linting GitHub Actions workflow works, it was distracting and reducing productivity for pull requests. So we decided to remove the workflow, and do occasional code styling chores instead, which is perfectly acceptable.

PR #387, opened 2024-05-03
Save full alpha as an attribute of the summary table

### LINKED ISSUE #384: Footnotes from `as_gt` when the total alpha is < 0.025
The following one-sided design spends a total alpha of 0.0125.
```
x <- gs_design_ahr(
  alpha = 0.0125,
  upper = gs_spending_bound,
  upar = list(sf = gsDesign::sfLDOF, total_spend = 0.0125, param = NULL),
  lower = gs_b,
  lpar = rep(-Inf, 3)
)
```
It has a list output containing `input`, `enroll_rate`, `fail_rate`, `bound`, and `analysis`.
image

Rather than using the list output directly, users commonly employ the `summary()` function to generate a summarized table of the design details described above.
```
x |> summary()
```
image

As evident from the summary output provided above, we observe that the total spent alpha information is NOT included. However, this information is crucial for incorporating the above summary table into a gt table for the footnotes. To address this, we can
```
x |> summary() |> as_gt()
x |>summary() |> as_gt(alpha = 0.0125)
```
image
image

**Question:** is there anyway to get the full alpha information from the summary table to the gt table?

LittleBeannie:
Fixes #384

### DIFF
--- a/R/as_gt.R
+++ b/R/as_gt.R
@@ -198,11 +198,6 @@ as_gt.fixed_design <- function(x, title = NULL, footnote = NULL, ...) {
 #'   `attr` is a vector of string to specify the attributes of the footnotes,
 #'   for example, `c("colname", "title", "subtitle", "analysis", "spanner")`;
 #'   users can use the functions in the `gt` package to customize the table.
-#' @param full_alpha The full alpha used in the design, the default is 0.025.
-#    If the cumulative alpha for final analysis is less than the `full_alpha`
-#'   when the futility bound is non-binding, a footnote will be displayed, saying
-#'   the smaller value subtracts the probability of crossing a futility bound before
-#'   crossing an efficacy bound at a later analysis under the null hypothesis.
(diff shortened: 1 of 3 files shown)


R/as_gt.R
```diff
@@ -306,10 +301,10 @@ as_gt.gs_design <- function(
     display_bound = c("Efficacy", "Futility"),
     display_columns = NULL,
     display_inf_bound = FALSE,
-    full_alpha = 0.025,
```
jdblischak, line 309: Is there any potential situation where a user would want to be able to manually override this?

jdblischak (approved):
Good work with `attr()` and `attributes()`! I recommend bumping the dev version since this affects user facing code, but it's not a hard requirement

nanxstats:
(I changed the `ifelse()` to `if-else` as the latter makes more sense for this case per [style guide](https://style.tidyverse.org/syntax.html?q=ifelse#if-statements))

Issue #346, opened 2024-02-23
Revert back `create_arm()` workaround after bug fix in upstream


nanxstats:
## Context

There is an if-condition that needs to be improved to be less stringent in the upstream dependency npsurvSS. Detailed analysis documented in #344.

We merged the short term solution (copy the original implementation, modify that condition, and make it an internal function) in #345.

## Next steps

- [x] Send a PR to the upstream repo to fix the issue and ask for a CRAN release.

Once this is fixed upstream and made it to CRAN, we should

- [x] Revert the changes in PR 345 so we rely on the upstream implementation.
- [x] Add the minimum version required (`npsurvSS (>= x.y.z)`) to `DESCRIPTION`.

LittleBeannie:
PR to npsurvSS: https://github.com/godwinyyung/npsurvSS/pull/4

Timeline: CRAN release by 3/11/2024

nanxstats:
Update: npsurvSS 1.1.0 is on CRAN 2024-05-08.

nanxstats:
Fixed by #391

PR #391, opened 2024-05-12
enhance npsurvSS dependency 

elong0527:
per #346 

- revert change in #235
- require npsurvSS >= 1.1.0 in description file


LittleBeannie (approved):
Thank you, @elong0527 !

PR #402, opened 2024-05-23
correct `gs_design_rd`

LittleBeannie:


R/gs_design_rd.R
```diff
@@ -215,7 +232,7 @@ gs_design_rd <- function(p_c = tibble::tibble(stratum = "All", rate = .2),
   } else if (info_scale == "h1_info") {
     (y_gs %>% filter(bound == "upper", analysis == k))$info1 / x_fix$info1[1]
   } else if (info_scale == "h0_h1_info") {
-    (y_gs %>% filter(bound == "upper", analysis == k))$info1 / x_fix$info0[1]
+    (y_gs %>% filter(bound == "upper", analysis == k))$info / x_fix$info1[1]
```
jdblischak, line 235: Could we create a test that covers this section of the if-else branch?

LittleBeannie: Good suggestion! Then we need to call for independent testers for this. Let's get the tests ready in another PR.

jdblischak: While independent tests are great, I'll note that we already have a file for [developer tests of gs_design_rd()](https://github.com/Merck/gsDesign2/blob/main/tests/testthat/test-developer-gs_design_rd.R). In other words, I don't see why we can't have both.

It looks like the example you added to the documentation would make a great developer test. The example states that the power of `x` and `y` should be equal. We can explicitly test that with `expect_equal()`.

keaven (approved):

PR #408, opened 2024-06-06
407 gs update ahr when futility bound is fixed the function does not work

LittleBeannie:
The following codes provide a two-sided design example with fixed futility bound. Because the futility bounds are fixed, it is not updated at the time of analysis.
```
 x <- gs_design_ahr(   
   enroll_rate = ..., fail_rate = ...,
   alpha = ..., beta = .., ratio = ...,
   info_scale = "h0_info",
   info_frac = NULL, analysis_time = c(20, 36),
   upper = gs_spending_bound,
   upar = list(sf = sfLDOF, total_spend = alpha),
   test_upper = TRUE,
   lower = gs_b,
   lpar = c(-1, 0),
   test_lower = c(TRUE, FALSE),
   binding = FALSE) |> to_integer()


 gs_update_ahr(
   x = x,
   ustime = ...,
   lstime = ...,
   observed_data = ...)
```

### DIFF
--- a/R/gs_update_ahr.R
+++ b/R/gs_update_ahr.R
@@ -332,7 +332,7 @@ gs_update_ahr <- function(
       upar_update$timing <- ustime
     } else {
       upar_update$timing <- ustime
-      if(!identical(x$input$lower, gs_b)) {
+      if(is.list(x$input$lpar)) {
         lpar_update$timing <- lstime
       }
     }
--- a/R/gs_update_ahr.R
+++ b/R/gs_update_ahr.R
@@ -230,17 +230,24 @@ gs_update_ahr <- function(
   # Check if is efficacy only
   one_sided <- all(x$bound$bound == "upper")
 
+  # Check if futility bound is fixed
+  fixed_futility_bound <- identical(x$input$lower, gs_b)
+
   # Check inputs ----
   if (is.null(x)) {
     stop("gs_update_ahr(): please input the original design created either by gs_design_ahr or gs_power_ahr.")
   }
 
-  if (!("ahr" %in% class(x))) {
-    stop("gs_update_ahr(): the original design must be created either by gs_design_ahr or gs_power_ahr.")
+  if (!any((c("ahr", "wlr") %in% class(x)))) {
+    stop("gs_update_ahr(): the original design must be created either by gs_design_ahr, gs_power_ahr, gs_design_wlr, or gs_power_wlr.")
   }
 
   if (one_sided && !is.null(lstime)) {
     stop("gs_update_ahr(): lstime is not needed for one-sided design.")
+  }
+
+  if (fixed_futility_bound && !is.null(lstime)) {
+    stop("gs_update_ahr(): lstime is not needed for two-sided design with fixed futility bounds.")
   }
 
   # Get the updated alpha ----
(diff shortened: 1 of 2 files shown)


R/gs_update_ahr.R
```diff
@@ -230,19 +230,26 @@ gs_update_ahr <- function(
   # Check if is efficacy only
   one_sided <- all(x$bound$bound == "upper")
 
+  # Check if futility bound is fixed
+  fixed_futility_bound <- identical(x$input$lower, gs_b)
```
nanxstats, line 234: This line will evaluate `gsDesign2::gs_b()` and compare its **content** with `x$input$lower`. It should work ok assuming `x` is computed on the fly. It will return `FALSE` in the unlikely case where `x` was loaded from a file (such as saved by the app) saved before some changes are made in `gsDesign2::gs_b()`.

I'm not sure how to avoid this - checking whether the saved function's **name** is `gs_b()` might be a way, although that is not currently saved in the `x` object when saving this input argument.

I'd welcome thoughts from @jdblischak or @yihui if there are any canonical solutions.

nanxstats (approved):

R/gs_update_ahr.R
yihui: Checking if two functions are identical is unlikely to be robust, and sounds like a bad idea to me. That means even cosmetic changes will lead to `FALSE`, e.g., looking at `gs_b()`,

https://github.com/Merck/gsDesign2/blob/f20836472c1e341a73741129d70654faa3ec6a77/R/gs_b.R#L71-L77

personally I'd like to change it to a one-liner:

```r
gs_b <- function(par = NULL, k = NULL, ...) {
  if (is.null(k)) par else par[k]
}
```

If we make such changes in the future, and `x` is loaded from a previously saved file, `fixed_futility_bound` will be `FALSE`.

Instead of comparing functions, I'd suggest comparing returned values of functions instead. If that's not possible, I'd probably add a unique attribute to `gs_b()` so that we can reliably identify it, e.g.,

```r
attr(gs_b, 'INTERNAL_NAME') <- 'GSDESIGN2_GS_B'
```

Then we test

```r
identical(attr(x$input$lower, 'INTERNAL_NAME'), 'GSDESIGN2_GS_B')
```

but I feel this solution is ugly.

I don't understand what the error message below means (`lstime is not needed...`). My instinct is that the error could be deferred elsewhere and is perhaps too early here.

LittleBeannie: Hi @yihui , thanks for the useful suggestion! 


As for **directly comparing the returned values**, I guess we may NOT be able do that... 😢 Although `lower=gs_b` returns numerical numbers directly for comparison, the similar story may not occur to `lower=gs_spending_bound`. When we call the `lower=gs_spending_bound` argument in `gs_design_ahr`, it passes the value of `lower` to `gs_design_npe`, which is a internal function of `gs_design_ahr` for more complicated calculations. In this way, the `lower` argument will not directly return some numerical numbers. 

As for **assigning attributes**, I guess it may could serve as a temporary solution?

I agree with you that a more robust way to compare 2 functions is desirable. I will create an issue for this and merge this PR first. 

Hope we could solve this nuisance soon!

PR #415, opened 2024-06-11
Update documentation for `gs_b()`

### LINKED ISSUE #414: The function documentation for `gs_b()` needs updated
I noticed that the function documentation for `gs_b()` is outdated:

https://github.com/Merck/gsDesign2/blob/cb91f5d6e16317490c56b6b4a1c80602a4c0481f/R/gs_b.R#L21-L28

As rendered in the [pgkdown site](https://merck.github.io/gsDesign2/reference/gs_b.html):

image

Problems I noticed:

* `gs_prob()`, `gs_power_nph()`, and `gs_power_nph()` have all been renamed
* While the first argument accepts Z-values, it is not named `Z` but `par`
* `par[k]j` should be `par[k]`
* It begins to suggest other functions to use for boundaries that change based on the information, but then never names any

jdblischak:
Closes #414

Notes:

* `gs_prob_combo()` has the tag `@noRd`, so I can't link to it
* The switch from `_nph` to `_npe` was purely a guess. If that's not correct, please let me know what to rename them
* I assumed that `gs_spending_bound()` was the function being hinted at by the final sentence

### DIFF
--- a/R/gs_b.R
+++ b/R/gs_b.R
@@ -18,9 +18,10 @@
 
 #' Default boundary generation
 #'
-#' `gs_b()` is the simplest version of a function to be used with the
-#' `upper` and `lower` arguments in `gs_prob_combo()`,
-#' [`gs_power_npe()`] and [`gs_design_npe()`].
+#' `gs_b()` is the simplest version of a function to be used with the `upper`
+#' and `lower` arguments in [`gs_power_npe()`] and [`gs_design_npe()`] or the
+#' `upper_bound` and `lower_bound` arguments `gs_prob_combo()` and
+#' `pmvnorm_combo()`.
 #' It simply returns the vector of Z-values in the input vector `par` or,
 #' if `k` is specified, `par[k]` is returned.
 #' Note that if bounds need to change with changing information at analyses,
--- a/R/gs_b.R
+++ b/R/gs_b.R
@@ -19,13 +19,13 @@
 #' Default boundary generation
 #'
 #' `gs_b()` is the simplest version of a function to be used with the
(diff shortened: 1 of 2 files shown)


R/gs_b.R
```diff
@@ -19,13 +19,13 @@
 #' Default boundary generation
 #'
 #' `gs_b()` is the simplest version of a function to be used with the
-#' `upper` and `lower` arguments in `gs_prob()`,
-#' `gs_power_nph()` and `gs_design_nph()`;
-#' it simply returns the vector input in the input vector `Z` or,
-#' if `k` is specified, `par[k]j` is returned.
+#' `upper` and `lower` arguments in `gs_prob_combo()`,
```
LittleBeannie, line 22: In additional to `gs_prob_combo`, `gs_power_npe` and `gs_design_npe`, shall we add `pmvnorm_combo()`? https://github.com/Merck/gsDesign2/blob/main/R/utility_combo.R#L146

nanxstats (commented):
Please see Yujie's comments. This fix looks good to me 👍

R/gs_b.R
jdblischak: I added `pmvnorm_combo()` in https://github.com/Merck/gsDesign2/pull/415/commits/59d6b6fc6ce6c543df2395777df6d7fe9b651f7b

Also, I noticed that the function arguments are different for `gs_prob_combo()` and  `pmvnorm_combo()`. These functions use `lower_bound` and `upper_bound`, so I updated that too

LittleBeannie (approved):
Thanks for enhancing the documentation, @jdblischak !

PR #418, opened 2024-06-12
416 no futility test at gs power ahr

LittleBeannie:


DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: gsDesign2
 Title: Group Sequential Design with Non-Constant Effect
-Version: 1.1.2.9
+Version: 1.1.3.0
```
jdblischak, line 3: Are we planning a CRAN release soon? Otherwise I would recommend only bumping the dev version. These can be arbitrarily large numbers
```suggestion
Version: 1.1.2.10
```

R/gs_power_ahr.R
```diff
@@ -172,21 +172,27 @@ gs_power_ahr <- function(
...
+    lpar <- rep(-Inf, n_analysis)
+    lower <- gs_b
+  }
+
   # Check if user input the total spending for futility,
   # if they use spending function for futility
-  if (identical(lower, gs_spending_bound) && is.null(lpar$total_spend)) {
-    stop("gs_power_ahr: please input the total_spend to the futility spending function.")
+  if (two_sided && identical(lower, gs_spending_bound)) {
+    if (is.null(lpar$total_spend) && any(test_lower)) {
```
jdblischak, line 193: There are many `if`s and `&&`s in this PR. It would be nice to add some developer tests to ensure this is working as expected (and that it keeps working). If you can provide me some example code, I could convert these to developer tests (in a follow-up PR).

jdblischak (changes requested):

R/gs_power_ahr.R
LittleBeannie: Sure, I will provide some tests examples with you in a follow-up PR.

jdblischak (approved):

PR #420, opened 2024-06-12
419 add developer test for gs power ahr

LittleBeannie:


DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: gsDesign2
 Title: Group Sequential Design with Non-Constant Effect
-Version: 1.1.2.10
+Version: 1.1.2.11
```
jdblischak, line 3: FYI in the future there is no need to bump the version number when only adding tests, since these aren't user facing. It's fine to leave it for this PR though

jdblischak (approved):
Nice tests!

PR #436, opened 2024-07-09
434 update `gs_update_ahr` to allow users update bounds at IA when data after this IA is missing

LittleBeannie:


R/gs_update_ahr.R
```diff
@@ -193,9 +193,19 @@
 #'   lstime = ustime,
 #'   observed_data = list(observed_data_ia, observed_data_fa))
 #'
-#' # alpha is upadted to 0.05
+#' # Example B5 ----
+#' # alpha is upadted to 0.05 ----
```
jdblischak, line 197: upadted -> updated (you'll also need to re-run roxygen2 to update the `.Rd` file)

jdblischak (approved):

PR #440, opened 2024-07-24
Update navbar style for pkgdown 2.1.0

nanxstats:
This PR updates the navbar link style in `extra.css` to accommodate the HTML class name changes in pkgdown 2.1.0 (due to the new option to have light/dark themes).

This would restore the original navbar look and feel under pkgdown < 2.1.0. Without this change, when rendering with pkgdown 2.1.0, the navbar link hover and active state color contrast will be too low.

### DIFF
--- a/pkgdown/extra.css
+++ b/pkgdown/extra.css
@@ -1,6 +1,5 @@
 /* navbar background */
-.bg-light,
-.navbar-light {
+nav.bg-light {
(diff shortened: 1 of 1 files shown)


jdblischak (approved):

LittleBeannie (approved):
Thanks, @nanxstats !

PR #444, opened 2024-08-05
The HSD spending function does not work for `gs_spending_combo()`

LittleBeannie:


R/gs_spending_combo.R
```diff
@@ -30,9 +30,77 @@
...
+#'
+#' par <- list(sf = gsDesign::sfXG3, total_spend = 0.025, param = 0.013)
+#' gs_spending_combo(par, info = 1:3 / 3)
+#'
 #' # beta-spending
 #' par <- list(sf = gsDesign::sfLDOF, total_spend = 0.2)
 #' gs_spending_combo(par, info = 1:3 / 3)
 gs_spending_combo <- function(par = NULL, info = NULL, ...) {
-  par$sf(par$total_spend, info, ...)$spend
+  par$sf(alpha = par$total_spend, t = info, param = par$param, ...)$spend
```
nanxstats, line 105: From [gsDesign technical manual](https://keaven.github.io/gsd-tech-manual/spendfn.html#sfArgs):

> ...a spending function passed to `gsDesign()` through the arguments (upper bound) and (lower bound) must have the following calling sequence: `sfname(alpha, t, param)`

and it looks like so for all existing spending functions in gsDesign. So maybe `...` is now redundant.

LittleBeannie: Thanks for connecting to gsDesign, @nanxstats. I deleted the `....` and it works.

R/gs_spending_combo.R
```diff
@@ -30,9 +29,77 @@
...
+#'
+#' # Xi and Gallo conditional error spending functions
+#' par <- list(sf = gsDesign::sfXG1, total_spend = 0.025, param = 0.5)
+#' gs_spending_combo(par, info = 1:3 / 3)
+#'
+#' par <- list(sf = gsDesign::sfXG2, total_spend = 0.025, param = 0.14)
+#' gs_spending_combo(par, info = 1:3 / 3)
+#'
+#' par <- list(sf = gsDesign::sfXG3, total_spend = 0.025, param = 0.013)
+#' gs_spending_combo(par, info = 1:3 / 3)
```
jdblischak, line 98: It wouldn't be a bad idea to add these examples to the tests. Currently there is only a single test in `test-independent-gs_spending_combo.R`. Can be done later though

jdblischak (approved):

nanxstats (approved):

PR #445, opened 2024-08-06
fix #442 by calculating upper bound based on spending function

### LINKED ISSUE #442: Inflated Type 1 Error in MaxCombo
Fixed design is using a fixed boundary for the calculation that needs to be calculated adaptively. 

https://github.com/Merck/gsDesign2/blob/main/R/fixed_design_maxcombo.R#L105

```
upper = gs_b, upar = qnorm(1 - alpha),
```

elong0527:
The issue discussed in #442 should be fixed after calculating boundary using spending function instead of a fixed boundary.


LittleBeannie (approved):
Thank you, @elong0527 !

PR #448, opened 2024-08-08
Tweaking to_integer()

yihui:
Removed the large amount of code repetition and the unnecessary `for`-loops. Please see individual commit messages if you find it difficult to understand the full changes.

I did this only because "while I'm looking at the code here..." :)

If my performance is evaluated based on the number of lines of code deleted rather than added, I can probably do better. I compared the `ahr` and `wlr` cases carefully by eyes, and it will be helpful if someone else could give it another pass to make sure I didn't miss other differences in the code for the two cases. First let's see if the PR would pass R CMD check.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: gsDesign2
 Title: Group Sequential Design with Non-Constant Effect
-Version: 1.1.2.18
+Version: 1.1.2.19
 Authors@R: c(
     person("Keaven", "Anderson", email = "keaven_anderson@merck.com", role = c("aut")),
     person("Yilong", "Zhang", email = "elong0527@gmail.com", role = c("aut")),
--- a/R/pw_info.R
+++ b/R/pw_info.R
@@ -239,7 +239,7 @@ pw_info <- function(
   ans <- ans[order(t), .SD, by = .(time, stratum)]
 
   # filter out the rows with 0 events
-  ans <- ans[!is_almost_k(event, 0L)]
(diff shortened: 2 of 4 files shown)


jdblischak (approved):

LittleBeannie (approved):
Thank you so much, @yihui! I gained valuable insights from  https://github.com/yihui/gsDesign2/blob/to_integer_tweaks/R/to_integer.R#L349 in order to avoid code duplication.

yihui:
@LittleBeannie You are welcome! Just FYI, `do.call()` can be handy sometimes, but it does have one drawback, i.e., when errors occur, the error message and traceback won't be as clear as when you call the function directly (there's no function name in the traceback).

```
f = function() stop('No')

f()
#> Error in f() : No

do.call(f, list())
#> Error in (function ()  : No
```

So normally I'd avoid using `do.call()` unless I need to call functions dynamically (e.g., conditionally) or call a function with dynamic arguments.

LittleBeannie:
> @LittleBeannie You are welcome! Just FYI, `do.call()` can be handy sometimes, but it does have one drawback, i.e., when errors occur, the error message and traceback won't be as clear as when you call the function directly (there's no function name in the traceback).
> 
> ```
> f = function() stop('No')
> 
> f()
> #> Error in f() : No
> 
> do.call(f, list())
> #> Error in (function ()  : No
> ```
> 
> So normally I'd avoid using `do.call()` unless I need to call functions dynamically (e.g., conditionally) or call a function with dynamic arguments.

Gained another valuable lesson, and thank you for sharing your experience with `do.call()`.

Issue #392, opened 2024-05-14
Hide the input parameters from a design object


LittleBeannie:
The current design object generated by `gs_design_ahr`, `gs_power_ahr`, etc., includes a lengthy list of input parameters that users may not find of interest. To hide these input parameters, I have incorporated them as design attributes.

nanxstats:
@jdblischak We should stick to S3. The summary return objects should not have the same class as the design objects - I'm talking about the critical one for method dispatching. If you look at `lm()` and `summary.lm()`:

```r
fit <- lm(1 ~ 1)
class(fit)
#> [1] "lm"
class(summary(fit))
#> [1] "summary.lm"
```

So the solution is simple: revisit how those classes got assigned and correct them.

jdblischak:
> I think one issue we are running into is that S3 is single dispatch. In other words, it can only choose a method based on a single class. It would be nice if we could have a summary method for objects with class fixed_design and ahr, but that's not possible with S3. We'd have to migrate to S4 to enable [multiple dispatch](https://adv-r.hadley.nz/s4.html#multiple-dispatch) on multiple classes, but S4 is a big jump in complexity and strictness that I doubt we want to take on.

I realized this was incorrect, so I want to clarify. Single versus multiple dispatch refers how many objects can influence the method dispatch. For example, in the hypothetical function call `f(a, b)`, if `f()` is an S3 generic, then it is only influenced by the class of `a`. But if it were an S4 generic, then it could be influenced by the class of both `a` and `b`.

With that out of the way, back to the strategy of assigning multiple classes to a single object.

```R
class(gsDesign2::gs_design_ahr())
## [1] "non_binding" "ahr"         "gs_design"   "list"
```

Only a single classed method will be dispatched, so why assign multiple classes? Because that way you can fall back to use methods for later classes. For example, you can define `summary.gs_design()` to create a custom summary (as is done in {gsDesign2}), but still take advantage of all the existing S3 methods defined for the class `list`:

```R
methods(class = "list")
##  [1] all.equal     as.data.frame as_tibble     cbind2        coerce        kronecker    
##  [7] Ops           rbind2        relist        type.convert  within       
## see '?methods' for accessing help and source code
```

Looking at the package `NAMESPACE`, we only have S3 methods assigned to two classes: `fixed_design` and `gs_design`

https://github.com/Merck/gsDesign2/blob/48a165be8acae2647299797bd47640862da8f8ea/NAMESPACE#L3-L10

[...]

Issue #282, opened 2023-08-18
Tidy up `summary` and `as_gt` functions


LittleBeannie:
The summary and as_gt functions in their current state are extremely lengthy. A preferable approach would be to condense them into a shorter function that utilizes sub-functions where necessary.

LittleBeannie:
Hi @jdblischak , considering we have couple of PRs to improve `as_gt` and `as_rtf` (such as https://github.com/Merck/gsDesign2/pull/458, https://github.com/Merck/gsDesign2/pull/450, https://github.com/Merck/gsDesign2/pull/449), can we close this issue? I can create another issue to tidy-up the nested if-else in `summary.gs_design` in another issue, to make the issue more specific.

yihui:
Or I can create a PR to clean up the summary functions, and close this issue once that PR is merged.

PR #449, opened 2024-08-09
Tweaking as_gt()

yihui:
I'm looking at the R scripts by alphabetical order, and reduced ~180 lines of code in `as_gt()`. Again, the changes may be easier to understand if you read the commits one by one.

In some places, I'm using my personal style (e.g., for line breaks and indentation), and happy to follow the style that other colleagues are comfortable with. In general, I tend to make the code a little compact so I can scroll less when reading the code.

I also realized `as_rtf()`'s source code is very similar to `as_gt()`, and perhaps can cut out more lines there lines after factoring out the common parts.

### DIFF
--- a/R/ahr_blinded.R
+++ b/R/ahr_blinded.R
@@ -84,14 +84,14 @@ ahr_blinded <- function(
     hr = c(1, .6),
     ratio = 1) {
   # Input checking
-  if (!is.vector(hr, mode = "numeric") || min(hr) <= 0) {
-    stop("ahr_blinded: hr must be a vector of positive numbers.")
+  if (!is.numeric(hr) || min(hr) <= 0) {
+    stop("'hr' must be a vector of positive numbers.")
   }
-  if (!is.vector(intervals, mode = "numeric") || min(intervals) <= 0) {
-    stop("ahr_blinded: intervals must be a vector of positive numbers.")
+  if (!is.numeric(intervals) || min(intervals) <= 0) {
+    stop("'intervals' must be a vector of positive numbers.")
   }
   if (length(intervals) != length(hr)) {
-    stop("ahr_blinded: the piecewise model specified hr and intervals are not aligned.")
+    stop("the piecewise model specified 'hr' and 'intervals' differ in lengths.")
   }
 
   # Set final element of "intervals" to Inf
@@ -111,11 +111,10 @@ ahr_blinded <- function(
   # Compute adjustment for information
   q_e <- ratio / (1 + ratio)
 
-  ans <- tibble(
+  tibble(
     event = sum(event),
     ahr = exp(-theta),
     theta = theta,
     info0 = sum(event) * (1 - q_e) * q_e
   )
-  return(ans)
 }
--- a/R/ahr_blinded.R
+++ b/R/ahr_blinded.R
@@ -91,7 +91,7 @@ ahr_blinded <- function(
     stop("'intervals' must be a vector of positive numbers.")
   }
   if (length(intervals) != length(hr)) {
-    stop("the piecewise model specified 'hr' and 'intervals' differ in lengths.")
+    stop("the piecewise model specified 'hr' and 'intervals' differ in length.")
   }
 
   # Set final element of "intervals" to Inf
--- a/R/ahr_blinded.R
+++ b/R/ahr_blinded.R
@@ -111,10 +111,11 @@ ahr_blinded <- function(
   # Compute adjustment for information
   q_e <- ratio / (1 + ratio)
 
-  tibble(
+  ans <- tibble(
     event = sum(event),
     ahr = exp(-theta),
     theta = theta,
     info0 = sum(event) * (1 - q_e) * q_e
   )
+  return(ans)
 }
--- a/R/as_gt.R
+++ b/R/as_gt.R
@@ -80,26 +80,9 @@ as_gt <- function(x, ...) {
 #'   as_gt()
 as_gt.fixed_design <- function(x, title = NULL, footnote = NULL, ...) {
   # get the design method
-  if ("ahr" %in% class(x)) {
-    design_mtd <- "ahr"
-  } else if ("fh" %in% class(x)) {
-    design_mtd <- "fh"
-  } else if ("mb" %in% class(x)) {
-    design_mtd <- "mb"
-  } else if ("lf" %in% class(x)) {
-    design_mtd <- "lf"
-  } else if ("rd" %in% class(x)) {
-    design_mtd <- "rd"
-  } else if ("maxcombo" %in% class(x)) {
-    design_mtd <- "maxcombo"
-  } else if ("milestone" %in% class(x)) {
-    design_mtd <- "milestone"
-  } else if ("rmst" %in% class(x)) {
-    design_mtd <- "rmst"
-  } else if ("rd" %in% class(x)) {
-    design_mtd <- "rd"
-  }
-
+  design_mtd <- intersect(
+    c("ahr", "fh", "mb", "lf", "rd", "maxcombo", "milestone", "rmst"), class(x)
+  )[1]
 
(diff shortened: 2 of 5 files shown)


jdblischak:
This is related to https://github.com/Merck/gsDesign2/issues/282

I still think that we eventually need to clean up the classes and methods as I noted in https://github.com/Merck/gsDesign2/issues/392. However, using `switch()` for now is definitely a great improvement in readability.

R/ahr_blinded.R
```diff
@@ -111,11 +111,10 @@ ahr_blinded <- function(
...
   q_e <- ratio / (1 + ratio)
 
-  ans <- tibble(
+  tibble(
     event = sum(event),
     ahr = exp(-theta),
     theta = theta,
     info0 = sum(event) * (1 - q_e) * q_e
   )
-  return(ans)
```
jdblischak, line 120: Personally I like the explicit return even though it is optional. And checking `git blame`, I'm pretty sure they existed before I started editing the code base. I'm happy to go along with the consensus though

yihui: I don't have a strong opinion here and can definitely use an explicit `return()`. Personally I use `return()` only when I need to return early.

R/ahr_blinded.R
```diff
@@ -84,14 +84,14 @@ ahr_blinded <- function(
     hr = c(1, .6),
     ratio = 1) {
   # Input checking
-  if (!is.vector(hr, mode = "numeric") || min(hr) <= 0) {
-    stop("ahr_blinded: hr must be a vector of positive numbers.")
+  if (!is.numeric(hr) || min(hr) <= 0) {
+    stop("'hr' must be a vector of positive numbers.")
   }
-  if (!is.vector(intervals, mode = "numeric") || min(intervals) <= 0) {
-    stop("ahr_blinded: intervals must be a vector of positive numbers.")
```
LittleBeannie, line 91: Hi Yihui, I have a question regarding the error message output. Typically, our practice in gsDesign2 is to print the function name at the beginning of the error message. This consideration arises from the possibility that the `ahr_blinded` function could potentially be invoked within another function. So... I am wondering if we should keep the `ahr_blinded` here?

R/as_gt.R
```diff
@@ -303,178 +240,102 @@ as_gt.gs_design <- function(
...
+  if (any(i <- display_columns == "Probability"))
+    display_columns <- c(display_columns[!i], "Alternate hypothesis", "Null hypothesis")
   ## check if the `display_columns` are included in `x` output
-  if (sum(!(display_columns %in% names(x))) >= 1) {
-    stop("as_gt: the variable names in display_columns is not outputted in the summary_bound object!")
-  } else {
-    x <- x %>% dplyr::select(dplyr::all_of(display_columns))
-  }
+  if (!all(display_columns %in% names(x))) stop(
+    "not all variable names in 'display_columns' are in the summary_bound object!"
```
LittleBeannie, line 285: Echo with my previous comment, shall we add `as_gt` at the beginning of the error message?

LittleBeannie (commented):
Thank you, @yihui, for tweaking the long `as_gt` function! I only have 2 minor comments, regarding the error message. Could you please take a look?

R/ahr_blinded.R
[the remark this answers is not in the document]
yihui: Good question. I deleted the function name in the error message because I thought it would be redundant.

```r
f = function() foo()
foo = function() {
  stop('No!')
}
f()
#> Error in foo() : No!
```

```r
foo = function() {
  stop('foo: No!')
}
f()
#> Error in foo() : foo: No!
```

This relates to a comment I made after https://github.com/Merck/gsDesign2/pull/448#issuecomment-2278528575 That is, the function name will be missing if you `do.call` the function.

```r
foo = function() {
  stop('No!')
}
f = function() do.call(foo, list())
f()
#> Error in (function ()  : No!
```

I think this case is relatively rare. If it does happen, `traceback()` can be helpful to know the chain of function calls.

That said, again I don't have a strong opinion here. In general, my principle is that I'll follow the preference of the project maintainer (i.e. you in this case).

R/ahr_blinded.R
yihui: @LittleBeannie Do you have a preference here? As I said, I'm totally fine with reverting to the original style `ans <- ...; return(ans)`, even though both the assignment and the explicit `return()` look unnecessary to me.

LittleBeannie: Thank you for verifying, @yihui ! I personally prefer the explicit return, as it provides a clearer message for non-advanced R programmers.

R/ahr_blinded.R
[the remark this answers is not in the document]
LittleBeannie: Thank you for the clear explanation; I understand your intention. I am open to this. If we choose to remove the function name at the beginning of the error message, we should review the entire gsDesign2 package, as the previous practice of gsDesign2 is to include the function name right at the start (maybe a new issue and PR).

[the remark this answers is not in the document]
yihui: Okay, as I'm going through the whole codebase, I'll do this along the way.

R/ahr_blinded.R
yihui: Got it. I've restored `return()` in 90352a04ad6d367b65777fd3915b5dc1ba5827fb.

LittleBeannie (approved):

R/ahr_blinded.R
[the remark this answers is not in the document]
jdblischak: I searched the repository for how `ahr_blinded()` is used. From the package code, examples, and tests, `ahr_blinded()` is always called directly. In other words, the function itself is never passed as an argument to another function. Thus the fact that an error comes from `ahr_blinded()` should be clear from the default error message, as @yihui noted above.

To demonstrate, I added `stop("there was an error")` to `ahr_blinded()` and generated the error message below:

```R
devtools::load_all()
ahr_blinded(
  surv = survival::Surv(
    time = simtrial::ex2_delayed_effect$month,
    event = simtrial::ex2_delayed_effect$evntd
  ),
  intervals = c(4, 100),
  hr = c(1, .55),
  ratio = 1
)
## Error in ahr_blinded(surv = survival::Surv(time = simtrial::ex2_delayed_effect$month,  : 
##   there was an error
```

R/as_gt.R
```diff
@@ -80,105 +80,44 @@ as_gt <- function(x, ...) {
...
-    )
-  }
-
+  if (is.null(title)) title <- sprintf("Fixed Design under %s Method", switch(
+    design_mtd,
+    ahr = "AHR", fh = "Fleming-Harrington", mb = "Magirr-Burman",
+    lf = "Lachin and Foulkes", maxcombo = "MaxCombo", milestone = "Milestone",
+    rd = "Fixed Design of Risk Difference under Farrington-Manning",
+    rmst = "Restricted Mean Survival Time"
+  ))
```
jdblischak, line 94: This isn't equivalent to the `switch()` statement that was replaced, for example when `design_mtd <- "rd"`

```R
# before
[1] "Fixed Design of Risk Difference under Farrington-Manning Method"
# after
[1] "Fixed Design under Fixed Design of Risk Difference under Farrington-Manning Method"
```

jdblischak (changes requested):
Overall this is looking great. Please see the one `switch()` statement that needs adjusted.

A bigger picture question for @LittleBeannie and @nanxstats : are we still planning on occasionally running {styler} on the code in {gsDesign2}? The default {styler} format is significantly different than Yihui's preferred style. Below is an example `git diff` after running {styler} on this PR:

image

R/as_gt.R
yihui: Good catch. Fixed now. Thanks!

jdblischak (approved):

PR #450, opened 2024-08-12
Refactor `as_rtf()`

yihui:
Massive refactoring of `as_rtf()` to share code with `as_gt()`.


R/as_rtf.R
```diff
@@ -296,7 +296,7 @@ as_rtf.gs_design <- function(
   orientation <- match.arg(orientation)
 
   method <- intersect(class(x), c("ahr", "wlr", "combo", "rd"))[1]
-  x_alpha <- max((x %>% dplyr::filter(Bound == display_bound[1]))[["Null hypothesis"]])
+  x_alpha <- max(filter(x, Bound == display_bound[1])[[colname_spannersub[2]]])
```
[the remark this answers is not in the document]
yihui: Now I think we should do it the other way around, i.e., we should make `as_gt()` consistent with `as_rtf()` by hardcoding the string since it seems `"Null hypothesis"` is hardcoded in `x`. Correct me if I'm wrong.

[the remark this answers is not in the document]
yihui: Forget it. I think I was wrong. Changing it will break a test.

R/as_rtf.R
```diff
@@ -299,6 +299,7 @@ as_rtf.gs_design <- function(
   x_alpha <- max(filter(x, Bound == display_bound[1])[[colname_spannersub[2]]])
   x_non_binding <- inherits(x, "non_binding")
   x_k <- as.numeric(substr(x$Analysis, 11, 11))
+  if (!display_inf_bound) x <- filter(x, !is.infinite(Z))
```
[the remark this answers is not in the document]
yihui: I added this line for `as_rtf()`, otherwise this argument would be meaningless. This also makes it consistent with `as_gt()`.

R/as_gt.R
```diff
@@ -397,3 +360,51 @@ footnote_non_binding <- function(x_alpha, full_alpha) {
...
+  r <- "^Analysis: ([0-9]+).*"
+  i <- grepl(r, a)
+  k <- as.numeric(sub(r, '\\1', a[i]))
+  i[i] <- k == max(k)
+  i & x$Bound == bound
+}
+
+gs_design_info <- function(
+    x, title, subtitle, spannersub, footnote, bound, columns, inf_bound,
+    alpha_column = spannersub[2], transform = identity
```
jdblischak, line 377: Might be worth documenting the argument `transform` since it isn't as straight-forward as the others

jdblischak (approved):

R/as_gt.R
yihui: Absolutely. I should document these new functions. Thanks for the suggestion!

R/as_gt.R
```diff
@@ -242,115 +250,24 @@ as_gt.gs_design <- function(
     display_inf_bound = FALSE,
     ...) {
 
-  method <- intersect(class(x), c("ahr", "wlr", "combo", "rd"))[1]
-  full_alpha <- attr(x, "full_alpha")
```
yihui, line 246: @LittleBeannie I have a question regarding a difference between `as_gt()` and `as_rtf()` that I noticed while comparing the two: for `as_gt()`, we obtain `full_alpha` from the attribute of `x`, whereas for `as_rtf()`, `full_alpha` is its argument and not retrieved from `x`'s attributes. Is this expected? I'm not sure if you or @elong0527 is the right person to ask.

R/as_gt.R
```diff
@@ -79,23 +79,37 @@ as_gt <- function(x, ...) {
...
+    gt::tab_header(title = title %||% fixed_title(method)) %>%
+    gt::tab_footnote(
+      footnote = footnote %||% method_footnote(x, method),
+      locations = gt::cells_title(group = "title")
+    )
+  return(ans)
+}
+
+# get the fixed design method
+fixed_method <- function(x) {
```
LittleBeannie, line 93: Is it appropriate to refer to this as the 'fixed_design_method'? From my perspective, the term 'fixed method' implies that the method is fixed, which may have limited clinical significance.

Furthermore, I've observed utility functions for both fixed design (such as `fixed_method`, `fixed_title`, `method_footnote`), as well as group sequential design (such as `get_display_columns `, `footnote_content`, `footnote_non_binding`, `footnote_row`, `gs_design_info`). I am considering whether it would be beneficial to use distinct prefixes for these two sets of utilities to differentiate their purposes.

Finally, I am wondering if we should save the above 2 sets of utilities into a separate R file, and name it as `utility_as_gt`.

R/as_gt.R
```diff
@@ -79,23 +79,37 @@ as_gt <- function(x, ...) {
...
+  methods <- c("ahr", "fh", "mb", "lf", "rd", "maxcombo", "milestone", "rmst")
+  intersect(methods, class(x))[1]
+}
 
-  # set the default title
-  if (is.null(title)) title <- sprintf("Fixed Design %s Method", switch(
-    design_mtd,
+# get the default title
+fixed_title <- function(method) {
```
LittleBeannie, line 99: Echo with my previous comment, can we call this function as `fixed_design_title`?

R/as_gt.R
```diff
@@ -79,23 +79,37 @@ as_gt <- function(x, ...) {
...
     milestone = "under Milestone", rmst = "under Restricted Mean Survival Time",
     rd = "of Risk Difference under Farrington-Manning"
   ))
+}
 
-  # set the default footnote
-  if (is.null(footnote)) footnote <- switch(
-    design_mtd,
+# get the default footnote
+method_footnote <- function(x, method) {
```
LittleBeannie, line 110: I am wondering why the prefix of `fixed_` is missing... Is the `method_footnote` appliable for both fixed design and group sequential design?

R/as_gt.R
LittleBeannie: Hi @yihui , yes, I can answer the question. The `full_alpha` should be retrieved from `x`'s attributes. I will prepare a PR to get it there.

R/as_gt.R
```diff
@@ -397,3 +362,54 @@ footnote_non_binding <- function(x_alpha, full_alpha) {
...
+  a <- x$Analysis
+  r <- "^Analysis: ([0-9]+).*"
+  i <- grepl(r, a)
+  k <- as.numeric(sub(r, '\\1', a[i]))
+  i[i] <- k == max(k)
+  i & x$Bound == bound
+}
+
+# a list of information for `as_[gt|rtf].gs_design()` methods
+gs_design_info <- function(
```
LittleBeannie, line 379: I am hesitant to use `gs_design_info` as the name, as `info` in gsDesign2 carries a specific meaning – statistical information (1/variance). Can we consider an alternative name for it?

LittleBeannie (commented):
Hi @yihui, there are lots of work, thank you so much! I only have some very minor comments, regarding the naming of the utility functions.

R/as_gt.R
yihui: If you want to do it, please wait until this PR is merged, or I can just do it in this PR.

LittleBeannie: Thanks for the confirmation. I know there are many changes occurring in this pull request, so let's wait until this PR is merged.

R/as_gt.R
yihui: > I am considering whether it would be beneficial to use distinct prefixes for these two sets of utilities to differentiate their purposes.

I'm using the prefixes `fd_` and `gsd_` now, and am open to other suggestions. I tend not to use the prefix `gs_design_` because I noticed that we have had functions with this prefix but serve a very different purpose. I hope the abbreviations `fd` and `gsd` make sense. They are only for ourselves and users won't use them, so the naming is perhaps not critically important. Anyway, naming is really hard :)

> Finally, I am wondering if we should save the above 2 sets of utilities into a separate R file, and name it as `utility_as_gt`.

I don't have an opinion on that. I can certainly put these functions in `as_utils.R` if you prefer.

R/as_gt.R
yihui: I was simply being lazy. It was only for fixed design.

R/as_gt.R
yihui: The naming of this function is the hardest. I spent nearly half an hour on it but couldn't think of a better name. I ended up using `gsd_parts()`. Please feel free to suggestion a different one.

LittleBeannie: Agree, naming is always the hardest part. As long as we could get rid of the `info` as keywords, we should be good. The new name `gsd_parts` looks good to me.

R/as_gt.R
LittleBeannie: I like the `fd` and `gsd`, which are sufficiently clear for developers. I am also good for the current folder/file structure. I will get it merged. Thanks for all the work!

LittleBeannie (approved):

PR #452, opened 2024-08-13
Let `to_integer()` always produce integer `analysis$n` and `analysis$event`

yihui:
Following the discussion at https://github.com/Merck/gsDesign2/pull/447#issuecomment-2276216039, I suggest that we round `n` and `event` without testing near equality because in theory, `n` should be nearly equal to `round(n)` (same for `event`). My understanding could be wrong, though.


LittleBeannie (approved):
Thanks a lot, @yihui !

Issue #459, opened 2024-08-26
Bug of `pw_info`


LittleBeannie:
The following code does not work.

```

# Poisson mixture model from  
p_pm <- function(x, theta, lambda, lower_tail = FALSE) {
  exp(-theta * (1 - exp(-lambda * x)))
}
# Cure rate determines theta
theta <- -log(cure_rate_control)
# lambda parameter driven by both cure rate and interim time rfs_18_control
lambda <- -log((theta + log(rfs_18_control)) / theta) / 18

# Time points of interest in design rate
weeks <- c(0, 6, 30)
months <- c(weeks / (52 + 1.25/7), 1:7) * 12

fail_rate_control = s2pwe(times = months[-1], survival = p_pm(months[-1], theta, lambda))
fail_rate <- tibble(stratum = "All", 
                    duration = fail_rate_control$duration, 
                    fail_rate = fail_rate_control$rate,
                    hr = c(1, 0.7, rep(.5, 7)),
                    # 0.6% dropout rate per month
                    dropout_rate = -log(.994))


pw_info(
  enroll_rate = tibble(stratum = "All", duration = 22, rate = 1),
  fail_rate = fail_rate,
  ratio = 2)
```

PR #460, opened 2024-08-26
Fix the bug of `pw_info` when there are many piecewise HRs

LittleBeannie:


LittleBeannie:
> Could the example in #459 be converted into an automated test?

Good idea, will get it added in a new commit!

keaven (approved):

jdblischak (approved):

PR #461, opened 2024-08-30
Clean up `summary` functions

### LINKED ISSUE #282: Tidy up `summary` and `as_gt` functions

yihui:
Closes #282.


R/summary.R
```diff
@@ -321,431 +284,139 @@ summary.gs_design <- function(object,
...
-        0.025
-      } else {
-        object$input$upar$total_spend
-      }
-    } else {
-      # case when given power to calcuate sample size
-      object$input$alpha
-    }
+# get a named vector of decimals (names are variable names)
+get_decimals <- function(vars, decs, vars_default, decs_default) {
```
yihui, line 386: This function should be equivalent to https://github.com/Merck/gsDesign2/blob/7a6ebb05c260926e781e127315d08cbf03449ab9/R/summary.R#L428-L462 but my code may be a little too abstract to understand.

yihui (commented):
I had to take several breaks while working on this PR, since it almost burned my brain several times. Sorry I can't make it [bite-sized](https://yihui.org/en/2018/02/bite-sized-pull-requests/) this time. Good luck to reviewers :)

R/summary.R
```diff
@@ -321,431 +284,139 @@ summary.gs_design <- function(object,
...
+  names(decs_default) <- vars_default
+  # Merge user-provided named decimals into default
+  decs_vars <- names(decs)
+  decs_default[decs_vars] <- decs
+
+  vars_name <- as.character(substitute(vars))
+  decs_name <- as.character(substitute(decs))
```
jdblischak, line 393: Could you please explain why it is necessary to run `substitute()` here? It isn't immediately obvious to me

R/summary.R
jdblischak: My code was very "brute force", so I'm fine with it being replaced with your more elegant function. I wrote a ton of tests for the various combinations of user input arguments, and they all pass, so I am confident it is equivalent.

jdblischak (changes requested):
No CI jobs are displayed for the PR because your most recent commit contained `[ci skip]`. The most recent run was this [one triggered by the previous push](https://github.com/Merck/gsDesign2/actions/runs/10714028405). I ran `R CMD check` locally with the latest version, and I can confirm there is still only the single `NOTE` about `setNames`.

R/summary.R
yihui: I can't praise the tests in this package enough. I have been quite impressed by their coverage.

R/summary.R
```diff
@@ -321,431 +284,139 @@ summary.gs_design <- function(object,
...
 
-  return(output)
+# capitalize variable names
+cap_names <- function(x) {
+  low <- c(
+    "analysis", "design", "power", "time", "event", "n", "bound", "z",
+    "~risk difference at bound", "~hr at bound", "~whr at bound", "nominal p"
+  )
+  # map lowercase names to capitalized names
+  map <- setNames(cap_initial(low), low)
```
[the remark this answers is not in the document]
yihui: Fixed now.

R/summary.R
yihui: Sure. I have added some `#` comments. Basically it was a trick to get the argument names `analysis_vars`, `analysis_decimals`, `col_vars`, and `col_decimals`, which will be used in the error messages below. Without the original argument names, the error messages will be less clear. We want an error message like this:

```
'col_decimals' must be a named vector if 'col_vars' is not provided
```

instead of

```
'decs' must be a named vector if 'vars' is not provided
```

jdblischak: Ah, thanks for the explanation. I see now that this was purely for the error message, which in retrospect makes total sense.

jdblischak (approved):
Amazing work. Thanks @yihui!

LittleBeannie (approved):
There are lots of work, thank you, @yihui !

R/summary.R
```diff
@@ -79,66 +79,29 @@
...
-  ans <- ans %>% dplyr::rename(Design = design)
+  # capitalize names with special case of Event -> Events
+  ans <- replace_names(ans, c(event = "Events")) %>% cap_names()
 
-  if ("n" %in% names(ans)) {
-    ans <- ans %>% dplyr::rename(N = n)
-  }
-
-  if ("event" %in% names(ans)) {
-    ans <- ans %>% dplyr::rename(Events = event)
```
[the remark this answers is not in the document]
LittleBeannie: Please use "Events" for both fixed design and group sequential designs.

R/summary.R
```diff
@@ -321,431 +284,139 @@ summary.gs_design <- function(object,
...
-    if ("~hr at bound" %in% names(xy)) {
-      bound_summary_detail <- xy %>% dplyr::rename("~whr at bound" = "~hr at bound")
-    } else {
-      bound_summary_detail <- xy
-    }
-  }
+  # If the method is WLR, change HR to wHR
+  if (method == "wlr") xy <- replace_names(xy, c("~hr at bound" = "~whr at bound"))
 
   # If the method is COMBO, remove the column of "~HR at bound", and remove AHR from header
```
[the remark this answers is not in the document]
LittleBeannie: The comment is correct, when it is maxcombo test, the "~HR at bound" and "AHR" do not exist.

PR #464, opened 2024-09-05
Import all functions once in `R/gsDesign2-package.R`, and remove the `pkg::` qualifiers

yihui:


LittleBeannie (approved):
Thanks for the comprehensive check and clean-up!

data-raw/simu_test_gs_design_combo.R
```diff
@@ -41,31 +41,31 @@ sim_gsd_pMaxCombo_exp1_H0 <- function(N = ceiling(454.60),
   failRates$failRate <- failRates$failRate * (failRates$hr + 1) / 2
   failRates$hr <- 1
 
-  failRates0 <- tibble::tibble(
+  failRates0 <- tibble(
```
jdblischak, line 44: This script is a bit of a mess. As it is I don't think it can even run. It never loads {dplyr} or {tibble} despite calling functions without the qualifier. It also calls a function `tensurv()`, which I couldn't find. This is a standalone script, not part of the R package, so we can't assume that the functions in `NAMESPACE` are available to it. If that is the behavior we want, we should add `devtools::load_all()` to the top of it.

nanxstats:
Apparently this is more of a personal preference and style guide thing so I'm afraid I can't be very helpful here.

The _R Packages_ book [dependencies chapter](https://r-pkgs.org/dependencies-in-practice.html#sec-dependencies-in-imports-r-code) currently pushes a style that qualifies external namespaces with `::`, without adding to `NAMESPACE` by default, unless it's used "a lot" or an operator. However, I don't think it's **strictly enforced** in their packages.

R/gsDesign2-package.R
```diff
@@ -16,8 +16,16 @@
...
-#' @keywords internal
-#' @useDynLib gsDesign2, .registration = TRUE
-#' @importFrom stats setNames
+#' @importFrom data.table ":=" as.data.table copy first last rbindlist setDF
+#'   setDT setcolorder setorderv
+#' @importFrom dplyr "%>%" all_of arrange desc filter full_join group_by lead
+#'   left_join mutate one_of rename row_number select summarize ungroup
+#' @importFrom gsDesign gsDesign sfLDOF
+#' @importFrom mvtnorm GenzBretz
+#' @importFrom stats pnorm qnorm setNames stepfun uniroot
```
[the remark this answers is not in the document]
yihui: That's also my preference. I'll revert the changes for `GenzBretz`.

data-raw/simu_test_gs_design_combo.R
yihui: The change in this script was accidental. I'll revert.

yihui:
I don't really have a strong opinion on this. My original motivation was that I saw sometimes we use `dplyr::`, and sometimes we don't, so I wanted to clean up the unnecessary `dplyr::` qualifiers. Then I moved on to remove other package qualifiers.

jdblischak:
> The _R Packages_ book [dependencies chapter](https://r-pkgs.org/dependencies-in-practice.html#sec-dependencies-in-imports-r-code) 

Later in that section, they also support defining `@importFrom` either separately for each function or in a central location. Thus I'm fine with migrating to the central location style.

jdblischak (approved):

nanxstats (approved):

PR #465, opened 2024-09-06
Follow-up on cleaning up `summary()` functions

yihui:
Two more changes according to https://github.com/Merck/gsDesign2/pull/461#discussion_r1747471816

### DIFF
--- a/R/summary.R
+++ b/R/summary.R
@@ -329,10 +329,6 @@ summary.gs_design <- function(object,
   # If the method is WLR, change HR to wHR
   if (method == "wlr") xy <- replace_names(xy, c("~hr at bound" = "~whr at bound"))
 
-  # If the method is COMBO, remove the column of "~HR at bound", and remove AHR from header
-  if (method == "combo" && "~hr at bound" %in% names(xy))
-    stop("'~hr at bound' can't be displayed!")
-
   output <- table_ab(
     # A data frame to be show as the summary header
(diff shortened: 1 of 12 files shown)


R/summary.R
```diff
@@ -332,10 +331,6 @@ summary.gs_design <- function(object,
   # If the method is WLR, change HR to wHR
   if (method == "wlr") xy <- replace_names(xy, c("~hr at bound" = "~whr at bound"))
 
-  # If the method is COMBO, remove the column of "~HR at bound", and remove AHR from header
-  if (method == "combo" && "~hr at bound" %in% names(xy))
-    stop("'~hr at bound' can't be displayed!")
-
```
yihui, line 338: Since `~hr at bound` can't possibly exist in `xy`, perhaps we can remove this check. It doesn't break any test.

@LittleBeannie I'm not sure if this is the right thing to do.

LittleBeannie: Hi @yihui , yes, you are right that `~hr at bound` does not exist in `xy`, because it is not part of the raw output from the MaxCombo test (see screenshot below).

Thanks for the confirmation!
image

yihui: Okay, then I guess we don't really need the check here. Thanks for letting me know!

LittleBeannie (approved):
Thank you, @yihui ! I will get it merged now.

PR #471, opened 2024-10-14
Improve `gs_cp()` documentation and run roxygen2

nanxstats:
This patch improves the `gs_cp()` documentation so that the format follows convention, exports it, and runs roxygen2.

Note that the target branch is `406-conditional-power`, not `main`.


R/gs_cp.R
```diff
@@ -16,27 +16,32 @@
...
+#' group sequential design
 #'
-#' @examples
-#' library(gsDesign2)
-#' library(dplyr)
+#' Computes conditional boundary crossing probabilities at future planned
+#' analyses for a given group sequential design assuming an interim
+#' z-statistic at a specified interim analysis.
+#'
+#' @param x An object of type `gs_design`.
```
LittleBeannie, line 26: Thanks, @nanxstats! One minor comments. The `x` is usually an output from `gs_update_ahr`.

As the `gs_cp` is still in progress, there are many upcoming changes to the documentation. If it's acceptable to you, could we complete the development first and then refine the documentation?

nanxstats: Okay! I'll close this for now.

PR #475, opened 2024-10-24
Link functions with roxygen2 Markdown syntax

nanxstats:
As title suggests. The [syntax](https://roxygen2.r-lib.org/articles/markdown.html#links) for linking functions in roxygen2 markdown is: `[func()]` and `[pkg::func()]`.

(Is it okay to change these now? Or should I wait until 2025? I just wanted to make sure I respect the priority of things and not standing in the way if other changes need to happen first.)


LittleBeannie (approved):
Thanks, @nanxstats !

PR #480, opened 2024-10-28
477 release gsdesign2 v113

LittleBeannie:
This PR will be merged after the PR #479  is merged.

PR #446, opened 2024-08-06
Info frac driven design does not work for `gs_design_wlr()`

LittleBeannie:
https://github.com/Merck/simtrial/issues/259

PR #481, opened 2024-10-29
Fix fixed_design_lf() example

jdblischak:
I fixed a minor typo in the example for `fixed_design_lf()` (it was calling `fixed_design_fh()`).

While checking this update, I noticed that `R CMD check` produced a NOTE about the variables `delta` and `sigma2` in `gs_design_wlr()`. I believe these were added a few weeks ago in https://github.com/Merck/gsDesign2/commit/f9aab1a217c44c3b7a58c196cfd185555513968a ( https://github.com/Merck/gsDesign2/pull/446). Given that we are preparing for a CRAN release (#480), I also fixed this.

### DIFF
--- a/R/fixed_design_lf.R
+++ b/R/fixed_design_lf.R
@@ -46,7 +46,7 @@
 #' x %>% summary()
 #'
 #' # Example 2: given sample size and compute power
-#' x <- fixed_design_fh(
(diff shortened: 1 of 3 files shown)


LittleBeannie (approved):
Nice catch, thank you @jdblischak!

PR #484, opened 2024-10-31
Update the rounding of events in `to_integer()`

LittleBeannie:
The current version of `to_integer()` floor the IA events and ceiling the FA events. In the PR, the IA events is rounded, and FA events is rounded up, as default.


keaven (approved):

keaven:
Yujie to delete old branch when ready.

PR #499, opened 2025-02-04
Allow `event_tbl` as input of `gs_update_ahr`

LittleBeannie:
When updating the boundaries, there are two approaches to consider:

- **Approach 1**: Input the blinded data. The current `gs_update_ahr` function will calculate the events for each piecewise interval. These events are then used to calculate the treatment effect (`theta`) and the statistical information under the alternative hypothesis (`info0`).
- **Approach 2**: Input the events for each piecewise interval directly.

Approach 2 is particularly useful when investigating boundaries with event accrual that is faster or slower than planned, without the need for observed data. Additionally, it is beneficial when users already know the summary of event counts.


R/gs_update_ahr.R
```diff
@@ -153,83 +83,48 @@
...
-#' # Updated design due to potential change of multiplicity graph
+#' # Alpha is updated to 0.05
 #' gs_update_ahr(x = x, alpha = 0.05)
 gs_update_ahr <- function(
     x = NULL,
     alpha = NULL,
     ustime = NULL,
     lstime = NULL,
-    observed_data = NULL) {
+    event_tbl = NULL) {
```
jdblischak, line 127: When I recommended only adding `event_tbl`, I didn't know that `observed_data` was already an option. Is this a common use case? I worry about backwards compatibility.

One option would be to start by supporting both `observed_data` and `event_tbl`, but emit a deprecation notice if a user provides `observed_data`. Then in a future release, we could remove `observed_data` completely to simplify the code.

jdblischak: Something like the following:

```R
if (!is.null(observed_data)) {
  warning(
    "The argument observed_data is deprecated and will be removed in a future release.\n",
    "Please update your code to use the argument event_tbl."
  )
  event_tbl <- func_to_convert(observed_data)
}
```

jdblischak: But this is ultimately your call. I don't know how common it was to use `observed_data`

LittleBeannie: Thank you for your suggestions! In some situations, we may use the entire TTE dataset as `observed_data`. However, I've noticed that using only `event_tbl` without `observed_data` seems to be more prevalent in practice.

Considering this, I decided to focus on only taking`event_tbl`. Additionally, I will create a function to convert `observed_data` into `event_tbl`, ensuring that we maintain flexibility in data handling.

Please let me know if you have any further thoughts or suggestions on this approach!

jdblischak (approved):

PR #501, opened 2025-02-18
improve the documentation of `gs_design_ahr` and `gs_power_ahr`

LittleBeannie:
Update the `@param`, `@return` and delete the `@section` in `gs_design_ahr` and `gs_power_ahr`.

### DIFF
--- a/R/gs_design_ahr.R
+++ b/R/gs_design_ahr.R
@@ -23,13 +23,13 @@
 #' @param ratio Experimental:Control randomization ratio.
 #' @param alpha One-sided Type I error.
 #' @param beta Type II error.
-#' @param info_frac Targeted information fraction at each analysis.
-#' @param analysis_time Minimum time of analysis.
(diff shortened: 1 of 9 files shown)


LittleBeannie:
Hi @jdblischak, I updated the documentation of `gs_design_ahr` and `gs_power_ahr`. Please review and check if everything is clearer now.

jdblischak (approved):
This is great! The difference between `gs_design_ahr()` and `gs_power_ahr()` is much clearer now, and I appreciate the new details about the return object.

My only request is that the key distinction in the Description ("Calculate sample size and bounds given the power" for `gs_design_X()` and "Calculate the power given the sample size" for `gs_power_X()`) be propagated to all the `gs_design_X()` and `gs_power_X()` functions. My initial confusion arose when comparing `gs_design_wlr()` and `gs_power_wlr()`, and the Description for these functions in unchanged. You can address this request in this PR or in a  a follow-up one.

PR #497, opened 2025-01-31
Refactor methods by passing table metadata as attributes

jdblischak:
This is an alternative approach to #482 to refactor the classes/methods as discussed in #457

This PR refactors the `fixed_design_X()` functions to define their own `design_display`, `title`, and `footnote` as attributes, which are then used by `summary()`, `as_gt()`, and `as_rtf()`. This idea was proposed by @yihui, which I described in https://github.com/Merck/gsDesign2/pull/482#issuecomment-2450559105

By using the above strategy, I was able to make the following improvements to the classes/methods:

* `summary.fixed_design()` no longer adds a method class to its output object (eg `"ahr"`) since this is no longer necessary. This was only used to constructed the table metadata, which is now already defined in the attributes of the object produced the respective `fixed_design_X()` function
* `summary.fixed_design()` returns an object of class `"fixed_design_summary"` to distinguish it from the output of a `fixed_design_X()` function (with class `"fixed_design"`)

Below is example code to test out the changes in this PR:

```R
devtools::load_all(".")

# ahr
x <- fixed_design_ahr(
  alpha = .025,
  enroll_rate = define_enroll_rate(duration = 18, rate = 20),
  fail_rate = define_fail_rate(
    duration = c(4, 100),
    fail_rate = log(2) / 12,
    hr = c(1, .6),
    dropout_rate = .001
  ),
  study_duration = 36
)
attributes(x)
## $names
## [1] "input"       "enroll_rate" "fail_rate"   "analysis"    "design"
##
## $design_display
## [1] "Average hazard ratio"
##
## $title
## [1] "Fixed Design under AHR Method"
##
## $footnote
## [1] "Power computed with average hazard ratio method."
##
## $class
## [1] "fixed_design" "list"
xsum <- summary(x)
attributes(xsum)
## $row.names
## [1] 1
##
## $names
## [1] "Design" "N"      "Events" "Time"   "Bound"  "alpha"  "Power"
##
## $class
## [1] "fixed_design_summary" "tbl_df"               "tbl"                  "data.frame"
##
## $title
## [1] "Fixed Design under AHR Method"
##
## $footnote
## [1] "Power computed with average hazard ratio method."
as_gt(xsum)
as_gt(xsum, title = "Custom title", footnote = "Custom footnote")
as_rtf(xsum, file = "refactor.rtf")
as_rtf(xsum, file = "refactor.rtf", title = "Custom title", footnote = "Custom footnote")

# fh
x <- fixed_design_fh(
  alpha = .025,
[...]

Issue #396, opened 2024-05-16
Update `lower` and `lpar` of `gs_power_ahr` to spending bounds with user specified `total_spend`


LittleBeannie:
The current version of `gs_power_ahr` use -Inf futility bounds. To be consistent with `gs_design_ahr`, we will update it to spending bounds with LDOF spending functions. However, the `gs_power_ahr` does not incoprate the value of `beta`, which makes it difficult to prespecify the `total_spend`. Considering this, we will ask user to prove the value of `total_spend`.

PR #502, opened 2025-02-19
Fix as_gt() examples

jdblischak:
The `as_gt()` examples are skipped during `R CMD check` since they generate a lot of extraneous output (#334, #338). However, this means they are not regularly tested. While working on #497, I discovered that the examples with `gs_power_ahr()` and `gs_power_wlr()` were broken.

```R
Error in gs_power_wlr() : 
  gs_power_wlr: please input the total_spend to the futility spending function.
```

I did a `git bisect` to determine that the example broke in https://github.com/Merck/gsDesign2/commit/a6a01660a5c6f32ad859b03c2b11583017bdd0da when we updated `gs_power_ahr()` and `gs_power_wlr()` to require that users specify `total_spend` (#396, https://github.com/Merck/gsDesign2/pull/398).

I fixed the examples as was done for `as_rtf()` in #398 and also updated the examples to be run during `R CMD check` to prevent this from happening again. The downsides are that `gsDesign2.Rcheck/gsDesign2-Ex.Rout` gets bloated with all the `as_gt()` output, and the examples are also decently long-running. If CRAN complains about either of these on the next submission, then we can re-evaluate the strategy. An alternative would be to copy the examples into a test file, and then return to skipping the examples during `R CMD check`.

### DIFF
--- a/R/as_gt.R
+++ b/R/as_gt.R
@@ -157,6 +157,7 @@ fd_footnote <- function(x, method) {
 #' @export
 #'
 #' @examplesIf !identical(Sys.getenv("IN_PKGDOWN"), "true")
+#' \donttest{
 #' # Group sequential design examples ---
 #'
 #' library(dplyr)
@@ -242,7 +243,7 @@ fd_footnote <- function(x, method) {
 #' gs_power_wlr(lpar = list(sf = gsDesign::sfLDOF, total_spend = 0.1)) %>%
 #'   summary() %>%
 #'   as_gt(display_columns = c("Analysis", "Bound", "Nominal p", "Z", "Probability"))
-#'
+#' }
 as_gt.gs_design <- function(
     x,
     title = NULL,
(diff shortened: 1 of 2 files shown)


nanxstats (changes requested):
Looking great, thanks for identifying this problematic behavior and the neat correction.

One small issue: CRAN will almost surely flag this check note as a problem, considering their build machine has even lower single-core performance than mine:

```text
* checking examples ... [27s/15s] NOTE
Examples with CPU (user + system) or elapsed time > 5s
       user system elapsed
as_gt 8.552  0.254   4.681
```

You probably know that the standard approach to avoid this note is by [wrapping the examples in `\donttest{}`](https://github.com/DavisVaughan/extrachecks?tab=readme-ov-file#you-get-told-not-to-comment-out-code-in-your-examples-section) - although they will still be ran.

jdblischak:
@nanxstats I have wrapped the `as_gt.gs_design()` examples in `\donttest{}`, and now `as_gt()` does not appear in any of the GitHub Actions `R CMD check` results.

There are some other example sections that occasionally run over 5s in some of the GitHub Actions jobs, but those weren't affected by my PR. I propose we wait and clean up any remaining longer-running examples in one-go prior to the next release (or the next time they are updated).

nanxstats (approved):
Thanks! Looking good to me.

PR #504, opened 2025-03-05
Assign the modified `df` back to `df` in `define_fail_rate()`

yihui:
It seems we forgot to assign the modified `df` back.


LittleBeannie (approved):
Thank you for catching it, @yihui !

nanxstats:
It looks like you have the same problem in `define_enroll_rate()` where `check_enroll_rate()` is used.

PR #507, opened 2025-03-06
Assign the modified enroll rate back

yihui:
similar fix as #504


LittleBeannie (approved):
Nice catch, thanks, @yihui !

PR #509, opened 2025-03-17
Allow bound functions and spending functions to take character values

yihui:
This will make it much easier to design the Shiny app.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: gsDesign2
 Title: Group Sequential Design with Non-Constant Effect
-Version: 1.1.3
+Version: 1.1.3.1
 Authors@R: c(
     person("Keaven", "Anderson", email = "keaven_anderson@merck.com", role = c("aut")),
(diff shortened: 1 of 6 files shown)


jdblischak:
xref: https://github.com/Merck/gsDesign2/pull/408#discussion_r1631815951, https://github.com/Merck/gsDesign2/issues/411, https://github.com/Merck/gsDesign2/pull/413

yihui:
Yes, a nice side-effect of this PR is that we will be able to compare functions robustly. The character name will be resolved to function _at run time_, and compared to a target function that is also retrieved at run time. This avoids cases like saving `gs_b` with an older version of **gsDesign2** and loading/comparing it against `gs_b` in a newer **gsDesign2** version some months later.

I managed to delete a large object of 75 lines in the Shiny app after this PR.

@LittleBeannie I think this PR is ready to merge now. I'm not sure the jargons in the news entry are accurate. Please feel free to tweak the words if necessary.

LittleBeannie (approved):
Thank you, @yihui! This looks good to me!

PR #511, opened 2025-03-20
Add back recently deleted SOP specs

jdblischak:
* The SOP specs for `gs_design_ahr()` and `gs_power_ahr()` were deleted in #501 
* Turns out these specs are required "to follow organization SOP for software development life cycle requirement" (https://github.com/Merck/gsDesign2/pull/508#issuecomment-2739975217)
* We may try to develop a system to better integrate these specs into the source code while still adhering to the SOP requirements (https://github.com/Merck/gsDesign2/pull/508#issuecomment-2740903348), but for now I simply wanted to add them back
* Note that there are extra changes in the `man/` directory because of recent changes to the roxgyen2 comments


LittleBeannie (approved):
Thank you for checking the compliance with the SOP, @jdblischak!

PR #526, opened 2025-04-10
Generate textual summary as gsDesign

LittleBeannie:
To fix issue #525.

## Textual summary of 1-sided design:
```
> x <- gs_design_ahr(info_frac = 1:3/3, test_lower = FALSE)
> x |> text_summary()
[1] "One-sided group sequential design with 3 analyses, time-to-event outcome with sample size 489.9 and 300.4 events, 90 percent power, 2.5 percent (1-sided) Type I error to detect hazard ratio of 0.9 during the first 3 months and 0.6 thereafter. Enrollment and total study durations are assumed to be 14 and 36 months, respectively. Efficacy bounds derived using a Lan-DeMets O'Brien-Fleming approximation spending function (no parameters)."

> x |> to_integer() |> text_summary()
[1] "One-sided group sequential design with 3 analyses, time-to-event outcome with sample size 490 and 301 events, 90 percent power, 2.5 percent (1-sided) Type I error to detect hazard ratio of 0.9 during the first 3 months and 0.6 thereafter. Enrollment and total study durations are assumed to be 14 and 36.1 months, respectively. Efficacy bounds derived using a Lan-DeMets O'Brien-Fleming approximation spending function (no parameters)."

> gs_power_ahr(event = c(10, 20, 30), test_lower = FALSE) |> text_summary()
[1] "One-sided group sequential design with 3 analyses, time-to-event outcome with sample size 108 and 30 events, 2.5 percent (1-sided) Type I error. Enrollment and total study durations are assumed to be 14 and 14.9 months, respectively. Efficacy bounds derived using a Lan-DeMets O'Brien-Fleming approximation spending function (no parameters). With hazard ratio of 0.9 during the first 3 months and 0.6 thereafter, the power is 9 percent."

> gs_power_ahr(event = c(10, 20, 30), test_lower = FALSE) |> to_integer() |> text_summary()
[1] "One-sided group sequential design with 3 analyses, time-to-event outcome with sample size 108 and 30 events, 2.5 percent (1-sided) Type I error. Enrollment and total study durations are assumed to be 14 and 14.9 months, respectively. Efficacy bounds derived using a Lan-DeMets O'Brien-Fleming approximation spending function (no parameters). With hazard ratio of 0.9 during the first 3 months and 0.6 thereafter, the power is 9 percent."
```

## Textual summary of a 2-sided symmetric design
```
> x <- gs_design_ahr(info_frac = 1:3/3,
+                    upper = gs_spending_bound, lower = gs_spending_bound,
[...]

PR #540, opened 2025-05-01
Use magrittr pipe with `text_summary()`

jdblischak:
The new function `text_summary()` uses the native pipe (https://github.com/Merck/gsDesign2/pull/526), which produces the following warning from `R CMD build`:

```
  NB: this package now depends on R (>= 4.1.0)
  WARNING: Added dependency on R >= 4.1.0 because package code uses the
  pipe |> or function shorthand \(...) syntax added in R 4.1.0.
  File(s) using such syntax:
    ‘text_summary.R’ ‘text_summary.Rd’
```

Currently our plan is to continue to support R >=3.5, so we need to use the magrittr pipe in the source code and examples (https://github.com/Merck/gsDesign2/pull/517)

### DIFF
--- a/R/text_summary.R
+++ b/R/text_summary.R
@@ -28,26 +28,26 @@
 #'
 #' # Text summary of a 1-sided design
 #' x <- gs_design_ahr(info_frac = 1:3/3, test_lower = FALSE)
-#' x |> text_summary()
-#' x |> to_integer() |> text_summary()
-#' gs_power_ahr(event = c(10, 20, 30), test_lower = FALSE) |> text_summary()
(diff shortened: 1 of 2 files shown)


nanxstats (approved):
Looking good. Once any of the `Imports` start to require R >= 4.1.0, we can bump it and switch to native pipes. So far all the `Imports` are all staying on >= 3.5 or so.

LittleBeannie (approved):
Thank you, @jdblischak, for catching this. This is an area I need to pay closer attention to in order to avoid future occurrences!

PR #541, opened 2025-05-02
Allow bound functions to be specified as character strings for `gs_power_ahr()`

yihui:
This is a very simple PR similar to #509, and needed for the Shiny app (for power calculation). It will be great to get it merged relatively soon. Thanks!


yihui:
Great! The power calculation in the app should be done shortly.

Issue #533, opened 2025-04-16
The design of the `weight = ...` argument in `gs_xxx_wlr`


LittleBeannie:
In the current versions of `gs_design_wlr` and `gs_power_wlr`, we set up the weights as follows.
```
weight = function(x, arm0, arm1) {wlr_weight_fh(x, arm0, arm1, rho = 0, gamma = 0.5)}
```

Here are some alternatives for setting up `weight = ...`.
```
weight = fh(0, 0.5)                         # mimic simtrial
weight = list(fun = fh, param =c(0, 0.5))   # mimic the spending function of gsDesign
```

The above alternatives may impact the summary of the weight function, which will be used during the textual summary of `gs_design_wlr` and `gs_power_wlr`.

LittleBeannie:
Decision: okay with the text input, like 
`weight = list(method = "fh", param = list(rho =0, gamma = 1))`
`weight = list(method = "mb", param = list(tau=Inf, w_max = 2))`
`weight = "logrank"`

PR #543, opened 2025-05-07
Change `weight` of a WLR design from a function to a list(weight = "fh", param = ...)

LittleBeannie:
To solve issue #533.

This issue lays the groundwork for the development of the weighting function summary, which is essential for generating a textual summary of a WLR design.


jdblischak (commented):
Great work @LittleBeannie! I like how this simplifies the the code

Issue #553, opened 2025-06-05
Fix the 2 NOTES from CRAN checking


LittleBeannie:
package gsDesign2_1.1.4.tar.gz does not pass the incoming checks automatically, please see the following pre-tests (additional issue checks):
Windows: <https://win-builder.r-project.org/incoming_pretest/gsDesign2_1.1.4_20250605_175723/Windows/00check.log>
Status: 2 NOTEs
Debian: <https://win-builder.r-project.org/incoming_pretest/gsDesign2_1.1.4_20250605_175723/Debian/00check.log>
Status: 1 NOTE

Last released version's CRAN status: OK: 3, NOTE: 10
See: <https://cran.r-project.org/web/checks/check_results_gsDesign2.html>

PR #554, opened 2025-06-05
Fix 2 NOTES of CRAN checking

LittleBeannie:
To solve issue #553.

### DIFF
--- a/R/text_summary.R
+++ b/R/text_summary.R
@@ -27,11 +27,8 @@
 #' library(gsDesign)
 #'
 #' # Text summary of a 1-sided design
-#' x <- gs_design_ahr(info_frac = 1:3/3, test_lower = FALSE)
+#' x <- gs_design_ahr(info_frac = 1:3/3, test_lower = FALSE) %>% to_integer()
 #' x %>% text_summary()
-#' x %>% to_integer() %>% text_summary()
-#' gs_power_ahr(event = c(10, 20, 30), test_lower = FALSE) %>% text_summary()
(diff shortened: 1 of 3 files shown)


R/text_summary.R
```diff
@@ -27,11 +27,8 @@
 #' library(gsDesign)
 #'
 #' # Text summary of a 1-sided design
-#' x <- gs_design_ahr(info_frac = 1:3/3, test_lower = FALSE)
+#' x <- gs_design_ahr(info_frac = 1:3/3, test_lower = FALSE) %>% to_integer()
 #' x %>% text_summary()
-#' x %>% to_integer() %>% text_summary()
-#' gs_power_ahr(event = c(10, 20, 30), test_lower = FALSE) %>% text_summary()
-#' gs_power_ahr(event = c(10, 20, 30), test_lower = FALSE) %>% to_integer() %>% text_summary()
```
jdblischak, line 34: If these examples would be informative for end users, you can wrap them in `\donttest{}` like I recently did for {simtrial} in https://github.com/Merck/simtrial/pull/332. It's your decision. Though if you do decide to use `\donttest{}`, you have to remember to re-document the package to update the `.Rd` file

jdblischak (approved):
Before you resubmit, I recommend submitting to the CRAN win-builder via `devtools::check_win_devel()`. This will let you know if all the NOTEs have been fixed

R/text_summary.R
LittleBeannie: Thanks for reminding me of the `\donottest{}`. After further consideration, I have decided to remove these examples, as they primarily demonstrate the implementation of `gs_power_ahr`, whether it is information-based or event-based. They do not effectively emphasize the use case of `text_summary`.

Issue #534, opened 2025-04-24
Correct spending in `gs_design_wlr`


LittleBeannie:
The following code generates a WLR design with 2 IAs and 1 FA. The alpha spending at IA1 should be `sfLDOF(alpha = 0.025, t = 0.2587722)$spend = 1.052152e-05` where 0.2587722 is the IA1 information fraction under the alternative hypothesis. However, the current `gs_design_wlr` outputs 8.714516e-06, which is the spending based on IA1 information fraction under the null hypothesis.

```
> x <- gs_design_wlr(analysis_time = 1:3*12, test_lower = FALSE)

# spending based on event fraction -- wrong!
> sfLDOF(alpha = .025, t = x$analysis$event / max(x$analysis$event))$spend[1] 
[1] 5.380432e-05

# spending based on information fraction under null -- wrong!
> sfLDOF(alpha = .025, t = x$analysis$info_frac0)$spend[1]
[1] 8.714516e-06

# spending based on information fraction under alternative hypothesis -- correct
> sfLDOF(alpha = .025, t = x$analysis$info_frac)$spend[1]
[1] 1.052152e-05

# the current IA1 spending is wrong!
> x$bound$`nominal p`[1]
[1] 8.714516e-06
```

LittleBeannie:
Following Keaven's publication of the spending time vignette at https://keaven.github.io/blog/delayed-effect-simulation/. I tried to reproduce the error of his MB example. With a quick check, the problem was caused by the `to_integer()` function.

## Version without `to_integer()`
```
> x <- gs_power_wlr(enroll_rate = define_enroll_rate(duration = 12, rate = 35.8),
+                   fail_rate = define_fail_rate(duration = c(4, 100),
+                                                fail_rate = log(2)/12,
+                                                dropout_rate = 0.001,
+                                                hr = c(1, 0.6)),
+                   analysis_time = c(20, 28, 36),
+                   weight = list(method = "mb", param = list(tau = NULL, w_max = 2)),
+                   upper = gs_spending_bound,
+                   upar = list(sf = gsDesign::sfLDOF, total_spend = 0.025),
+                   lpar = rep(-Inf, 3),
+                   lower = gs_b,
+                   test_lower = FALSE) 
> x|> summary() |> as_gt()

# the alpha-spending at the first analysis should be:
> sfLDOF(alpha = .025, t = x$analysis$info_frac0)$spend[1]
[1] 0.001445226
 
# rather than 
[...]

PR #557, opened 2025-06-24
Correct the IA1 crossing probability under H0 for WLR designs

LittleBeannie:
To solve issue #534.


keaven (approved):
Thanks for the update.

Issue #560, opened 2025-06-30
Add `h1_spending =...` to `gs_power_wlr`


LittleBeannie:
The current version of `gs_power_wlr` does not include` h1_spending = ...` argument. I will add this feature in an upcoming PR.

![Image](https://github.com/user-attachments/assets/4bb0d4fb-649b-4b1e-a469-c785f2644edc)

PR #561, opened 2025-06-30
Add the `h1_spending` argument to `gs_power_wlr`

LittleBeannie:
To solve issue #560.

### DIFF
--- a/R/gs_power_wlr.R
+++ b/R/gs_power_wlr.R
@@ -220,7 +220,9 @@ gs_power_wlr <- function(enroll_rate = define_enroll_rate(duration = c(2, 2, 10)
 
   # Given the above statistical information calculate the power ----
   y_h1 <- gs_power_npe(
-    theta = x$theta, theta0 = 0, theta1 = theta1,
+    theta = x$theta,
+    theta0 = 0,
+    theta1 = theta1,
     info = x$info,
     info0 = x$info0,
     info1 = info1,
@@ -237,7 +239,9 @@ gs_power_wlr <- function(enroll_rate = define_enroll_rate(duration = c(2, 2, 10)
   )
 
   y_h0 <- gs_power_npe(
-    theta = 0, theta0 = 0, theta1 = theta1,
+    theta = 0,
+    theta0 = 0,
+    theta1 = theta1,
     info = x$info0,
     info0 = x$info0,
     info1 = info1,


R/gs_power_wlr.R
```diff
@@ -208,12 +209,21 @@ gs_power_wlr <- function(enroll_rate = define_enroll_rate(duration = c(2, 2, 10)
...
+    info1 <- x$info
+  } else {
+    theta1 <- 0
+    info1 <- x$info0
+  }
+
   # Given the above statistical information calculate the power ----
   y_h1 <- gs_power_npe(
-    theta = x$theta,
+    theta = x$theta, theta0 = 0, theta1 = theta1,
```
jdblischak, line 223: Since all the other arguments are on their own line, I recommend keeping this consistent with the surrounding code and moving `theta0` and `theta1` to new lines

R/gs_power_wlr.R
```diff
@@ -227,12 +237,10 @@ gs_power_wlr <- function(enroll_rate = define_enroll_rate(duration = c(2, 2, 10)
   )
 
   y_h0 <- gs_power_npe(
-    theta = 0,
-    theta0 = 0,
-    theta1 = x$theta,
+    theta = 0, theta0 = 0, theta1 = theta1,
```
jdblischak, line 240: Ditto above comment

tests/testthat/test-developer-gs_power_wlr.R
```diff
@@ -0,0 +1,13 @@
+test_that("Validate 2-sided symetric design", {
+  x <- gs_power_wlr(enroll_rate = define_enroll_rate(duration = 12, rate = 50),
+                    analysis_time = NULL, event = c(100, 200, 300),
+                    upper = gs_spending_bound,
+                    upar = list(sf = gsDesign::sfLDOF, total_spend = 0.025),
+                    lower = gs_spending_bound,
+                    lpar = list(sf = gsDesign::sfLDOF, total_spend = 0.025),
+                    binding = TRUE, h1_spending = FALSE)
```
jdblischak, line 8: Love the test! For completeness, would it also make sense to have a test with `h1_spending = TRUE` to demonstrate that in that case the upper and lower bounds are not equal?

LittleBeannie:
This PR is closed since it is duplicated with PR #565. Keaven suggested testing issue #562 and #560 together in one branch.

Issue #562, opened 2025-07-22
`gs_design_wlr` does not approximate `gs_design_ahr` when it is logrank test


LittleBeannie:
The following lines of code generate an AHR design with 2 IAs and 1 FA.
```
enroll_rate <- define_enroll_rate(duration = c(2, 2, 2, 18), rate = c(1, 2, 3, 4), stratum = "All")

duration <- diff(c(0, 4, 6, 44))
control_rate <- log(2)/c(6, 6, 6)
fail_rate <- define_fail_rate(duration, fail_rate = control_rate, dropout_rate = c(0.001, 0.001, 0.001), hr = c(1, 0.8, 0.6), stratum = "All")

x1 <- gs_design_ahr(
  enroll_rate = enroll_rate,
  fail_rate = fail_rate,
  alpha = 0.025,
  beta = 0.1,
  analysis_time = c(9, 27, 36),
  info_frac = NULL,
  ratio = 1,
  binding = FALSE,
  upper = "gs_spending_bound",
  upar = list(sf = "sfLDOF", total_spend = 0.025, param = NULL),
  lower = "gs_spending_bound",
  lpar = list(sf = "sfLDOF", total_spend = 0.1, param = NULL),
  h1_spending = TRUE,
  info_scale = "h0_h1_info") 

x1 |> 
  summary(analysis_decimals = c(info_frac0 = 4)) |>
  gsDesign2::as_gt(
    title = "Asymmetric 2-sided design with beta-spending and non-binding futility",
    subtitle = "Both alpha- and beta-spending used",
    footnote = FALSE)
```
<img width="300" height="200" alt="Image" src="https://github.com/user-attachments/assets/20eec3e9-5e41-4bcf-9cc2-726adb8ced56" />

When we use `gs_design_wlr` to generate a WLR design under logrank test, the results should be approximately the same. However, it is not with IA1 missing. After some debugging work, I find the IA1 info0 is very small at the WLR design, leading to Inf efficacy bound. While this small IA1 info is not correctly calculated.   
```
x2 <- gs_design_wlr(
  enroll_rate = enroll_rate,
  fail_rate = fail_rate,
  alpha = 0.025,
  beta = 0.1,
  analysis_time = c(9, 27, 36),
  info_frac = NULL,
  ratio = 1,
  binding = FALSE,
  weight = "logrank",
  upper = "gs_spending_bound",
  upar = list(sf = "sfLDOF", total_spend = 0.025, param = NULL),
[...]

PR #563, opened 2025-07-23
Fix the issue where WLR design is different from AHR design even under logrank tests

LittleBeannie:
To solve issue #562.


keaven (approved):
Checks out when comparing to gs_design_ahr.

Issue #564, opened 2025-07-23
Update a vignette documentation


LittleBeannie:
`simfix` -> `sim_fixed_n`
https://github.com/Merck/gsDesign2/blob/v1.1.5/vignettes/articles/story-ahr-under-nph.Rmd#L423

PR #566, opened 2025-08-08
Update simtrial syntax in the vignette

LittleBeannie:
To solve issue #564.


jdblischak (changes requested):
I ran the vignette locally to confirm it gives similar results. The code changes look good. However, the vignette text still describes `simtrial::simfix()`. Less important, but the chunk labels also refer to `simfix`. To avoid confusion, I recommend removing all references to "simfix"

LittleBeannie:
> I ran the vignette locally to confirm it gives similar results. The code changes look good. However, the vignette text still describes `simtrial::simfix()`. Less important, but the chunk labels also refer to `simfix`. To avoid confusion, I recommend removing all references to "simfix"

Nice catch! I updated the text documentation and re-request your review.

Issue #520, opened 2025-04-03
Improve the help file of `gs_power_npe` and `gs_design_npe`


LittleBeannie:
This issue targets the documentation of `gs_power_npe` and `gs_design_npe`:

1. Improve the help file to provide a clear distinction between `info` and `info1`.
2. Combine the help files for `gs_power_npe` and `gs_design_npe`.

PR #567, opened 2025-08-25
Improve the help file of `gs_power_npe` and `gs_design_npe`

LittleBeannie:
To solve issue #520.


R/gs_power_npe.R
```diff
@@ -156,7 +177,8 @@
 #'   lpar = c(qnorm(.1), -Inf, -Inf)
 #' )
 #'
-#' # Spending function bounds
+#' # Example 8 ----
```
jdblischak, line 180: There are two "Example 8"

PR #577, opened 2025-08-29
Migrate completely to native pipe

jdblischak:
Close #576 

This PR migrates the entire package to use the native pipe. It removes the import of `%>%` from {dplyr} and also requires R >= 4.1.0.

I wasn't sure about the R files in `data-raw/` and `inst/` since these are more of one-off scripts documenting what was done. I went ahead and converted them too, but can revert if needed.


nanxstats (changes requested):
A few places have things like `@importFrom dplyr ... "|>"` which makes no sense - the `"%>%"` part should be simply removed instead of being replaced with `"|>"` because dplyr does not export `|>`.

LittleBeannie (approved):
Thanks, @jdblischak!

PR #582, opened 2025-09-09
Remove Rplots.pdf from test directory

jdblischak:
Minor PR. It can be included in 1.1.6 (#580, #581) or wait until post CRAN release.

xref: https://github.com/Merck/gsDesign2/pull/571#pullrequestreview-3156907525, #572


LittleBeannie (approved):
Thanks, I will include it to v1.1.6

PR #591, opened 2025-10-29
Fix assignment of attribute `uninteger_is_from` by `to_integer()`

jdblischak:
This is a follow-up to my big refactor of the S3 classes in https://github.com/Merck/gsDesign2/pull/584. I missed the use of `class(x)` at the end of `to_integer()` to assign the attribute `uninteger_is_from`. This caused the wrong value of `uninteger_is_from` to be assigned because the method is no longer embedded in the class hierarchy.

https://github.com/Merck/gsDesign2/blob/5103d9c21bba96a45d43be6545a4e65db30747b2/R/to_integer.R#L579-L592

While writing a test for this, I discovered that `to_integer()` is unable to process the object returned by `gs_power_rd()` because it doesn't have the element `input` with the values of its input arguments. I went ahead and fixed that too.

### DIFF
--- a/R/gs_power_rd.R
+++ b/R/gs_power_rd.R
@@ -343,9 +343,18 @@ gs_power_rd <- function(
       select(analysis, n, rd, rd0, theta1, theta0, info, info0, info_frac, info_frac0)
   )
 
+  # Get input parameters to output ----
+  input <- list(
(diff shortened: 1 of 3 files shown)


LittleBeannie (approved):
Thank you, @jdblischak! I got it merged.

PR #622, opened 2026-04-29
Remove "Time" from the updated design summary

### LINKED ISSUE #616: Remove "Time" from the updated design gt-summary table
When we present the updated design (`gs_update_ahr` object) summary, we will skip the reporting of "Time".

Image

LittleBeannie:
To solve issue #616

### DIFF
--- a/R/gs_update_ahr.R
+++ b/R/gs_update_ahr.R
@@ -394,7 +394,6 @@ gs_update_ahr <- function(
 
   ans$analysis <- data.frame(
     analysis = 1:n_analysis,
-    time = x$analysis$time,
     n = x$analysis$n,
     event = if (is.null(event_tbl)) {
       x$analysis$event
--- a/R/gs_bound_summary.R
+++ b/R/gs_bound_summary.R
@@ -82,16 +82,19 @@ gs_bound_summary_single <- function(x, col_efficacy_name = "Efficacy", digits,
     info <- paste0(info, "%")
     n <- round(analysis$n[i])
     events <- round(analysis$event[i])
-    time_value <- round(analysis$time[i], tdigits)
-    # If the value is an integer, force it to have the correct number of decimal places
-    time_value <- format(time_value, nsmall = tdigits)
+    time_value <- analysis$time[i]
+    if (!is.null(time_value)) {
+      time_value <- round(time_value, tdigits)
+      # If the value is an integer, force it to have the correct number of decimal places
+      time_value <- format(time_value, nsmall = tdigits)
+    }
 
     col_analysis <- c(
       col_analysis,
       if (i == final) "Final" else paste(label, info),
       paste("N:", n),
       paste("Events:", events),
-      paste0(timename, ": ", time_value),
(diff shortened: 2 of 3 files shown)


LittleBeannie:
Hi @jdblischak, we encountered an error in `gs_bound_summary`. Could you please help with that?

R/gs_update_ahr.R
```diff
@@ -394,7 +394,6 @@ gs_update_ahr <- function(
 
   ans$analysis <- data.frame(
     analysis = 1:n_analysis,
-    time = x$analysis$time,
```
jdblischak, line 397: This is a breaking change to remove `time` from the object output. If you want to remove it from the summary table, why not just update `summary.gs_design()` to remove that column?

jdblischak (commented):
The error `non-numeric argument to mathematical function` occurs for the line `time_value <- round(analysis$time[i], tdigits)` in `gs_bound_summary_single()` after running `gs_update_ahr()` with a new alpha value:
    https://github.com/Merck/gsDesign2/blob/24492c62527c74f8c2bb4141ce4663134185918b/R/gs_bound_summary.R#L85

This is because `time` was removed the data frame `analysis` returned by `gs_update_ahr()`.

Do you also want to remove the Time from the table produced by `gs_bound_summary()`? If not, the tests can be fixed by adding back `time` to the output from `gs_update_ahr()`.

R/gs_update_ahr.R
LittleBeannie: We are not typically advised to report the "Time" variable in the updated design output from `gs_update_ahr`. Therefore, I have removed it directly from the raw output at the source.

The motivation for this change comes from observing that statisticians often work directly with the raw output, bypassing the `summary()` function. Removing "Time" at the source ensures that the output is cleaner and more aligned with practical use cases, preventing potential misuse or confusion.

LittleBeannie:
> The error `non-numeric argument to mathematical function` occurs for the line `time_value <- round(analysis$time[i], tdigits)` in `gs_bound_summary_single()` after running `gs_update_ahr()` with a new alpha value:
> 
> https://github.com/Merck/gsDesign2/blob/24492c62527c74f8c2bb4141ce4663134185918b/R/gs_bound_summary.R#L85
> 
> This is because `time` was removed the data frame `analysis` returned by `gs_update_ahr()`.
> 
> Do you also want to remove the Time from the table produced by `gs_bound_summary()`? If not, the tests can be fixed by adding back `time` to the output from `gs_update_ahr()`.

"Time" is not reported for returns from `gs_update_ahr()`.
"Time" is reported for returns from `gs_design_xxx()`, `gs_power_xxx()`, and `fixed_design_xxx()`.

R/gs_update_ahr.R
jdblischak: Fixed in 1ff3537856477d3c3d99d23370edc74aaa628a27

jdblischak (approved):
I updated `gs_bound_summary()` in 1ff3537856477d3c3d99d23370edc74aaa628a27

PR #623, opened 2026-05-05
perf: optimize gs_design_ahr (~3-4x speedup)

yihui:
## Summary

- Optimize `gs_design_ahr()` and its dependency chain for 3-4x speedup across all input variants
- Replace expensive `object.size()` calls in cache pruning with O(1) `numhash()` check
- Replace dplyr operations (tibble, mutate, full_join, select, arrange, filter) with base R equivalents in hot-path functions (`gs_power_npe`, `gs_design_npe`, `gs_design_ahr`, `expected_time`, `gs_info_ahr`)
- Rewrite `expected_event()` internals using pure vector arithmetic instead of data.frame/merge/order operations (8x speedup for this function alone)

## Benchmark results (20 calls each, after warm-up)

| Scenario | Before | After | Speedup |
|----------|--------|-------|---------|
| Default (single analysis) | 0.985s | 0.327s | **3.0x** |
| Multiple analysis_time (3 analyses) | 2.707s | 0.997s | **2.7x** |
| Info_frac driven | 2.847s | 1.157s | **2.5x** |
| Info_frac + analysis_time | 3.790s | 1.318s | **2.9x** |
| 2-sided symmetric (O'Brien-Fleming) | 3.059s | 0.910s | **3.4x** |
| gs_b lower (no futility bound) | 3.399s | 0.773s | **4.4x** |

## Key changes by commit

1. **prune_hash**: Replace `object.size()` (walks entire hash, ~2ms/call) with `numhash()` (O(1)) for the frequent check; clear when entry count > 100
2. **gs_power_npe output**: Replace `tibble()` + `mutate()` + `arrange()` with `data.frame()` + base R sort (called 9+ times per design via gs_design_npe's root-finding)
3. **gs_design_npe output**: Replace `full_join` + `select` + `rename` + `arrange` with `merge()` + column subsetting
4. **gs_design_ahr output**: Replace dplyr chain (mutate, full_join, select, arrange, filter) with base R equivalents
5. **Hot-path functions**: Remove `dplyr::select()`, `dplyr::mutate()`, `dplyr::transmute()` from `expected_time`, `gs_info_ahr`, and the info_frac loop in `gs_design_ahr`
6. **expected_event**: Rewrite internals using vectors instead of data.frame/merge/order (8x speedup: 3.6ms to 0.43ms per call)
7. **Backward compatibility**: Exported functions (`gs_power_npe`, `gs_design_npe`) still return tibbles via `as_tibble()` at the return boundary

## Test plan

[...]

PR #624, opened 2026-05-05
Optimize return object preparation from gs_power_npe()

jdblischak:
As discussed in group meeting, this PR optimizes the preparation of the return object by `gs_power_npe()`.

Sorry @yihui. I had prepared this yesterday afternoon, but didn't have time to submit the PR. Could we review/merge this targeted PR first, and then rebase your more comprehensive PR on top?

xref: #623

### DIFF
--- a/R/gs_power_npe.R
+++ b/R/gs_power_npe.R
@@ -384,7 +384,7 @@ gs_power_npe <- function(theta = .1, theta0 = 0, theta1 = theta, # 3 theta
     }
   }
 
-  ans <- tibble(
+  ans <- data.frame(
     analysis = rep(1:n_analysis, 2),
     bound = c(rep("upper", n_analysis), rep("lower", n_analysis)),
     z = c(b, a),
@@ -392,14 +392,10 @@ gs_power_npe <- function(theta = .1, theta0 = 0, theta1 = theta, # 3 theta
     theta = rep(theta, 2),
     theta1 = rep(theta1, 2),
     info_frac = rep(info / max(info), 2),
-    info = rep(info, 2)
-  ) |>
-    mutate(
-      info0 = rep(info0, 2),
-      info1 = rep(info1, 2)
-    ) |>
-    # filter(abs(Z) < Inf) |>
-    arrange(desc(bound), analysis)
+    info = rep(info, 2),
+    info0 = rep(info0, 2),
+    info1 = rep(info1, 2)
+  )
 
   return(ans)
 }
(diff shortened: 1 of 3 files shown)


jdblischak:
Explanation of changes:

* `tibble()` -> `data.frame()` - no advantage to using tibble here, and we know from experience it is noticeably slower
* `mutate()` - unnecessary because `info0` and `info1` can be defined in the `data.frame()` call above
* `filter()` - removed this dead code. It was already commented out 4 years ago when the function was originally added in https://github.com/Merck/gsDesign2/commit/18337df4810adaf024bc74500878a180f4e09810 (#26)
* `arrange()` - removed. This sorting is unnecessary. By construction of the columns `analysis` and `bound`, the order will always be `upper` followed by `lower`, and the analyses in ascending order `(1:n_analysis)`

I updated the tests by running `as.data.frame()` on the output of the legacy implementation `gs_power_npe_()`. I felt this was the most transparent option. Another option would have been to ignore the attributes on the objects, but I felt this was more opaque (and more verbose), eg:

```R
  local_edition(3)
  expect_equal(x1_c, x2, ignore_attr = TRUE)
```

And lastly, we could continue to return a tibble via `as_tibble()` as Yihui did in #623. Just let me know if you prefer one of these alternative solutions.

jdblischak:
To benchmark the improvement, I put my changes in a separate `gs_power_npe2()` so that I could directly compare to the original. It reduced the median runtime from ~7 ms to ~4 ms. The most noticeable difference is in the worst case maximum time. This was reduced from ~41 ms to ~12 ms.

```R
microbenchmark::microbenchmark(
  before = gs_power_npe(
    theta = c(.1, .2, .3),
    info = (1:3) * 40,
    upper = gs_b,
    upar = c(Inf, 3, 2),
    lower = gs_b,
    lpar = c(qnorm(.1), -Inf, -Inf)
  ),
  after = gs_power_npe2(
    theta = c(.1, .2, .3),
    info = (1:3) * 40,
    upper = gs_b,
    upar = c(Inf, 3, 2),
    lower = gs_b,
    lpar = c(qnorm(.1), -Inf, -Inf)
  ),
  check = "equivalent"
)
## Unit: milliseconds
##    expr    min      lq     mean  median     uq     max neval cld
##  before 6.5527 6.81630 7.635802 7.14665 7.4633 41.6138   100  a 
##   after 4.0639 4.16815 4.518994 4.27110 4.5039 12.0563   100   b
```

yihui (approved):
No worries. I'm all for using data.frame() here! Thanks!

I simply asked Claude to optimize gs_design_ahr() without specific instructions yesterday, and it naturally chose base R. I hope it didn't do so just to accommodate my personal taste.

jdblischak:
> it naturally chose base R. I hope it didn't do so just to accommodate my personal taste.

No, I doubt it. Base R is clearly the right choice in this context. The data is too small to justify using something like {data.table}. For a simple gathering of results into a table, `data.frame()` is the way to go.

jdblischak:
@LittleBeannie After you approve, please squash and merge this PR. I don't think it justifies a NEWS bullet, but please let me know if you feel otherwise.

LittleBeannie (approved):
Thanks for the fix!

PR #625, opened 2026-05-06
Convert tests from testthat to testit

yihui:
## Summary

- Replace all testthat tests with testit equivalents (60 R test files + 2 markdown snapshot files)
- Snapshot tests (`as_gt`, `as_rtf`) converted to testit's markdown-based `.md` format with output embedded inline
- DESCRIPTION updated: `testthat (>= 3.0.0)` replaced with `testit`
- Key conversions: `test_that()` → `assert()`, `expect_equal()` → `all.equal()`, `expect_identical()` → `%==%`, `expect_error()` → `has_error()`, `expect_true()` → `(expr)`

## Motivation

```r
> # testthat dependencies
> setdiff(unlist(tools::package_dependencies('testthat', recursive = TRUE)), xfun::base_pkgs())
 [1] "brio"      "callr"     "cli"       "desc"      "evaluate"  "jsonlite"  "lifecycle"
 [8] "magrittr"  "pkgload"   "praise"    "processx"  "ps"        "R6"        "rlang"    
[15] "waldo"     "withr"     "fs"        "glue"      "pkgbuild"  "rprojroot" "diffobj"  
[22] "crayon"   

> # testit dependencies (none)
> setdiff(unlist(tools::package_dependencies('testit', recursive = TRUE)), xfun::base_pkgs())
character(0)
```

More info: https://yihui.org/en/2026/05/testthat-to-testit/


yihui:
@jdblischak @LittleBeannie This PR is ready for review.

tests/testit/helper.R
```diff
@@ -1,3 +1,15 @@
+gt_to_latex <- function(data) cat(as.character(gt::as_latex(data)))
+
+all_equal <- function(target, current, tolerance = 1e-5, ...) {
```
LittleBeannie: I noticed that this function is not exported from the testit package.

I was wondering about the design decision behind this. Could you please share the motivation for keeping it outside of `testit`?

tests/testit/test-developer-ahr.R
```diff
@@ -29,14 +29,13 @@ test_that("unstratified population, compared with old version", {
     dplyr::rename(time = Time, ahr = AHR, event = Events)
 
   # verify columns without n (sample size)
-  expect_equal(as.data.frame(y1), as.data.frame(x2))
+  (all_equal(as.data.frame(y1), as.data.frame(x2)))
```
LittleBeannie: Is there a redudant pair of parentheses?

yihui: No, that's testit's syntax for tests. You may read the PR description above.

tests/testit/helper.R
yihui: Please see advantages 1, 2, and 3 above in Part 2 of the PR description above. testit doesn't provide helper functions like all_equal() here. There is only one simple rule: you write tests in `()` that it should return TRUE to pass. Your tests can be any R code and testit is not opinionated about which functions, operators, or packages you use to write the test code (e.g., you can use `==`, `>`, `is.list()`, `all.equal()`, ...).

The reason we need `all_equal()` here is testthat's `expect_equal()` is too complicated and I had to provide a helper function like this to maintain compatibility with testthat's behavior. Since you asked, I think it will be a better idea to use all.equal(), or even better, just use `abs(target - current) <= tolerance`, which I think will be much clearer than both `testthat::expect_equal()` and `all.equal()`, but when I was trying this yesterday, I found we are not always comparing the absolute difference to tolerance in our tests, but sometimes absolute difference and sometimes relative difference, which brought this helper function `all_equal()`. I'll take a closer look today to see which equality tests rely on absolute differences and which rely on relative differences.

.github/workflows/R-CMD-check.yaml
```diff
@@ -39,11 +39,8 @@ jobs:
           http-user-agent: ${{ matrix.config.http-user-agent }}
           use-public-rspm: true
 
-      - uses: r-lib/actions/setup-r-dependencies@v2
+      - uses: yihui/actions/setup-r-dependencies@HEAD
```
jdblischak: I appreciate the desire for convenience, but pointing directly to `HEAD` makes me uneasy. I'd prefer using Git tags. And given the increasing prevalence of supply chain attacks via GitHub Actions, using [immutable releases](https://docs.github.com/en/code-security/concepts/supply-chain-security/immutable-releases) would be even safer.

tests/test-all.R
```diff
@@ -0,0 +1,2 @@
+library(testit)
+test_pkg("gsDesign2")
```
jdblischak, line 2: In order to use this new testing infrastructure in RStudio via `Ctrl/Cmd + Shift + T`, we'd need to remove the field `PackageUseDevtools: Yes` from `gsDesign2.Rproj`

jdblischak (commented):
I confirmed that the test coverage reported locally by `covr::package_coverage()` before and after is identical (93.12%) 🎉 

A few ergonomic questions:

* Why does {testit} need to be loaded in order to run `test_pkg()`? I couldn't find this in the docs

```R
packageVersion("testit")
## [1] ‘0.18.5’
testit::test_pkg()
## Error from sys.source2(r, envir = env, top.env = ns)
## Error in assert("unstratified population, compared with old version",  :
##   could not find function "assert"
```

* Is it possible to suppress the many error messages printed to the R console? I assume these are from tests that use `has_error()`

```R
library("testit")
test_pkg()
# Lots of error messages printed to the console, eg
## Error: missing value where TRUE/FALSE needed
## Error: missing value where TRUE/FALSE needed
## Error: `times` (c(1, 2, 1)) must be positive and strictly increasing!
## Error: `survival` (c("0.5", "NA")) must be positive!
## Error: `survival` must be of same length as `times`
## Error: `survival` (c(0.5, -0.1)) must be positive!
## Error: `survival` must be non-increasing
## Error: `survival` must be non-increasing

# But they all passed
.Last.value
## NULL
```

* I observed that {testit} stops on the first error. Is there any plan to enable {testit} to collect test errors to spot patterns, or is this out of scope?

* For quick feedback, I often use the argument `filter` of `devtools::test()`, eg `devtools::test(filter = "npe")`. Is it possible to do something similar with {testit}?

.github/workflows/R-CMD-check.yaml
yihui: Of course you are right. I'll tag them and pin to specific tags.

yihui:
@jdblischak All ergonomic issues have been addressed in testit. May thanks to all your great suggestion! I think they improved the usability of this package by 10 times.

jdblischak (commented):
Given the recent updates to {testit} (thanks @yihui!), I am supportive of this migration of the testing framework.

I observed a slight reduction in the code coverage locally via `covr::package_coverage()` (93.12% to 92.91%), but according to Codev the coverage is unchanged at [92.90%](https://app.codecov.io/github/Merck/gsDesign2/commit/cedcbef667c1ca53265b2083b7a34235e3081bcb).

I would like to delay merging until 1) we switch to tagged versions of `yihui/actions`, and 2) the updated version of {testit} is uploaded to CRAN.

yihui:
@jdblischak Actions have been tagged. The CRAN release of testit can be made at any time if you don't have further suggestions or requests.

yihui (commented):
This PR is ready for review. I'll release testit v1.0 to CRAN later today and switch to that version accordingly once it lands on CRAN.

To help reviewers understand the changes better, I need to clarify one extra thing I did in this PR, which I should have saved for another PR but I was not sure if it'd be worth torturing you one more time :) The extra thing was that I tightened the equality tests (i.e. `expect_equal()`) as much as possible for greater rigor. The logic is the following:

1. Switch to `identical()` (or equivalently, `%==%` from testit) whenever possible. If the two objects are strictly identical to each other, we use `%==%` instead of testing for approximate equality.
2. When we test things like `nrow()` or `ncol()` that are integers, we also try identical testing by changing the target to an integer, e.g., ``expect_equal(nrow(z1$`_footnotes`), 1)`` is changed to ``(nrow(z1$`_footnotes`) %==% 1L)`` (note the change from double `1` to integer `1L`); this is because `1` is not identical to `1L` although they are "equal".
3. Then for the rest of equality tests, try `all.equal()` with default tolerance (i.e. `sqrt(.Machine$double.eps)`). If that works, we just use the default `all.equal()`.
4. If the default tolerance is too tight, we raise it to the nearly smallest tolerance possible to make the test pass, e.g., if the actual difference is 0.00085, we use 0.001. In our original tests, tolerances (if explicitly provided) are often too high, e.g., when 0.003 works but we used 0.01.
5. When comparing probabilities, we usually use the argument `scale = 1` for testing the _absolute difference_ between two probabilities. This gives us a crystal clear idea about how much the two probabilities differ.

BTW, if test coverage is a concern, I think we can easily address it in the next PR.

I'll also add a skill to give AI models instructions on how to write tests with testit in future.

.claude/skills/write-tests.md
```diff
@@ -0,0 +1,196 @@
...
+  ...
+})
+```
+
+## Running Tests
+
+```bash
+cd tests
+Rscript test-all.R
+```
```
jdblischak, line 181: Is there any reason to prefer changing to the directory `tests/` versus running the tests from the project directory?

```sh
Rscript tests/test-all.R
```

.claude/skills/write-tests.md
```diff
@@ -0,0 +1,196 @@
...
+
+Run a subset of tests by passing a regex filter:
+
+```bash
+cd tests
+Rscript -e "testit::test_pkg('gsDesign2', filter = 'independent-ahr')"
```
jdblischak:
```suggestion
Rscript -e "library(testit); test_pkg('gsDesign2', filter = 'independent-ahr')"
```

It will fail with `Error: could not find function "assert"` if you do not load {testit} first.

.github/workflows/R-CMD-check.yaml
```diff
@@ -39,11 +39,8 @@ jobs:
           http-user-agent: ${{ matrix.config.http-user-agent }}
           use-public-rspm: true
 
-      - uses: r-lib/actions/setup-r-dependencies@v2
+      - uses: yihui/actions/setup-r-dependencies@v1.1.0
         with:
-          extra-packages: any::rcmdcheck
-          needs: check
+          r-universe: yihui cran
```
jdblischak: After the next CRAN release, can we remove the reliance on r-universe builds? If this serves an additional purpose (eg to obtain Linux binaries), then at the very least we should limit it to the `cran` r-universe.

DESCRIPTION
```diff
@@ -62,7 +62,7 @@ Suggests:
     knitr,
     rmarkdown,
     simtrial,
-    testthat (>= 3.0.0)
+    testit
```
jdblischak: Should we add a minimum required version? Given that we use snapshot testing, my guess is that we need at least version [0.16](https://github.com/yihui/testit/blob/master/NEWS.md#changes-in-testit-version-016)

jdblischak (commented):
The improvements to the tolerances are very welcome! These have long been a source of frustration for me.

Let's merge once the latest {testit} is available from CRAN.

.claude/skills/write-tests.md
yihui: No, no reason at all. `test_pkg()` works fine no matter if you are at the package root or under the tests/ directory. I just changed it to `Rscript tests/*.R`.

.claude/skills/write-tests.md
yihui: Oh right. I just added some command-line support in testit: https://github.com/yihui/testit/commit/03f9861fe040f99996abb667cbb98cbc197b4432 Now we can simple call `Rscript tests/*.R` with the `--filter` argument. I think that will make things simpler.

.github/workflows/R-CMD-check.yaml
yihui: Of course. The `r-universe` config will be removed here. The `cran` r-universe is used by default (for installing Linux binaries if available).

DESCRIPTION
yihui: Yes.

yihui: [testit 1.0](https://github.com/yihui/testit/releases/tag/v1.0) is on CRAN, and I've updated DESCRIPTION and GHA workflows accordingly.

jdblischak (approved):
🚀

jdblischak:
Just noticed this [NOTE](https://github.com/Merck/gsDesign2/actions/runs/25804483061/job/75803252363?pr=625#step:6:98). Could you please add `.claude` to `.Rbuildignore`?

```
* checking for hidden files and directories ... NOTE
Found the following hidden files and directories:
  .claude
These were most likely included in error. See section ‘Package
structure’ in the ‘Writing R Extensions’ manual.
```

yihui:
Sure. Done.

jdblischak:
Reminder to please squash and merge this PR

yihui:
I'm not sure who are admins of this repo, but they can go to https://github.com/Merck/gsDesign2/settings and disable "merge commits" and "rebase merging", which is what I do for all my repositories since I rarely need the full commit history of a PR and I always squash and merge:

image

PR #639, opened 2026-06-09
Unify test names for consistency

jdblischak:
I recently noticed that a minority of the test files follow the inconsistent filename pattern `test-independent_`, while the majority use either `test-independent-` or `test-developer-`. For consistency, I updated these files to instead use a dash. We only use an underscore when reflecting the name of the R function being tested.

The only judgement call was for `test-independent_wlr_weight.R` because `test-independent-wlr_weight.R` already existed, and both of these files have existed in parallel for as long as I could track (4 years ago in https://github.com/Merck/gsDesign2/commit/55804c5ce7f15863dc6992961b59e1d654bf0251). I decided to combine them, but I'm of course open to alternative solutions.

```sh
ls tests/testit/test-independent_* -1 | wc -l
## 9
ls tests/testit/test-independent-* -1 | wc -l
## 28
ls tests/testit/test-developer-* -1 | wc -l
## 28
ls tests/testit/test-developer_* -1 | wc -l
## 0

ls tests/testit/test-independent*wlr_weight.R -1
## tests/testit/test-independent_wlr_weight.R
## tests/testit/test-independent-wlr_weight.R
```


yihui (approved):
Sounds good to me!

PR #641, opened 2026-06-22
perf: rewrite expected_event internals using vectors

yihui:
## Summary

- Rewrite `expected_event()` internals using pure vector arithmetic instead of data.frame/merge/order operations
- Replace `stats::stepfun()` with `stepfun2()` (already on main) for rate lookups
- 8x speedup for this function (3.6ms → 0.43ms per call), ~2.5x for `pw_info`

Split from #623 to keep the `expected_event.R` optimization reviewable independently.

## Key changes

- Compute union of enrollment/failure breakpoints directly as vectors
- All intermediate arithmetic on plain numeric vectors (no data.frame allocation)
- Only construct data.frame for final output when `simple = FALSE`

## Test plan

- [x] `test-developer-expected_event.R` passes
- [x] `test-independent-expected_event.R` passes
- [x] Numerical output identical to baseline

### DIFF
--- a/R/expected_event.R
+++ b/R/expected_event.R
@@ -141,99 +141,86 @@ expected_event <- function(
   # Report error where there is >=2 strata
   if(length(unique(fail_rate$stratum)) >= 2 || length(unique(enroll_rate$stratum)) >= 2) {stop("Please calculate the expected event by stratum, see examples. ")}
 
-  # Divide the time line into sub-intervals ----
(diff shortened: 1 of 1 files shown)


yihui:
Closing due to concerns about potential breakage (a huge amount of effort has been made on this `expected_evant()` function and we'd avoid touching it if possible).

PR #642, opened 2026-06-22
perf: replace stats::stepfun with stepfun2 in expected_event

yihui:
## Summary

- Replace 4 `stats::stepfun()` calls with `stepfun2()` in `expected_event()`
- Improve `stepfun2()` itself: use a for-loop of vectorized comparisons (`x >= b`) for small breakpoint vectors (n <= 10), falling back to `findInterval` for larger ones

## Why

`stats::stepfun()` is **57x slower to create** than `stepfun2()` due to S3 class setup, environment construction, and method registration. In `expected_event`, step functions are created fresh on every call and evaluated once on small vectors (3-5 elements) — creation cost dominates.

The loop approach avoids `findInterval`'s dispatch overhead and binary search, which are overkill for 3-4 breakpoints. Each iteration is a single vectorized primitive (`x >= b`), with no allocation.

## Benchmark (the 4-stepfun create+eval pattern in expected_event, 50k calls)

| Implementation | Time | Speedup vs `stats::stepfun` |
|---|---|---|
| `stats::stepfun` | 10.8s | 1x |
| `stepfun2` (old, `findInterval`) | 1.0s | 11x |
| `stepfun2` (new, loop for n<=10) | 0.5s | **21x** |

## Test plan

- [x] `test-developer-expected_event.R` passes
- [x] `test-independent-expected_event.R` passes
- [x] Verified `stepfun2` produces identical results to `stats::stepfun` for both `right = FALSE` and `right = TRUE`

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -51,9 +51,13 @@ is_wholenumber <- function (x, tol = .Machine$double.eps^0.5)  {
 # a faster version of stats::stepfun() since we don't need to consider interpolation
 stepfun2 <- function(x0, y, right = FALSE) {
   x0; y  # avoid lazy evaluation: evaluate right now
-  function(x) {
-    i <- findInterval(x, x0, left.open = right)
-    y[i + 1]
+  if (length(x0) <= 10L) {
+    if (right)
+      function(x) { i <- 1L; for (b in x0) i <- i + (x > b); y[i] }
+    else
+      function(x) { i <- 1L; for (b in x0) i <- i + (x >= b); y[i] }
+  } else {
+    function(x) y[findInterval(x, x0, left.open = right) + 1L]
   }
 }
 
--- a/R/utils.R
+++ b/R/utils.R
@@ -48,9 +48,14 @@ is_wholenumber <- function (x, tol = .Machine$double.eps^0.5)  {
   abs(x - round(x)) < tol
 }
 
-# a faster version of stats::stepfun() since we don't need to consider interpolation
+# a faster version of stats::stepfun() since we don't need to consider interpolation;
+# right = FALSE means left-closed intervals [x0[i], x0[i+1]), i.e., the function jumps
+# at x0[i] (right-continuous); right = TRUE means right-closed (x0[i], x0[i+1]]
 stepfun2 <- function(x0, y, right = FALSE) {
   x0; y  # avoid lazy evaluation: evaluate right now
+  # for small breakpoint vectors, a for-loop of vectorized comparisons (x >= b)
+  # is faster than findInterval(), which has dispatch overhead and uses binary
+  # search (overkill for 3-5 breakpoints)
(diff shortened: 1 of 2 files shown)


R/utils.R
```diff
@@ -48,12 +48,21 @@ is_wholenumber <- function (x, tol = .Machine$double.eps^0.5)  {
...
-    i <- findInterval(x, x0, left.open = right)
-    y[i + 1]
+  # for small breakpoint vectors, a for-loop of vectorized comparisons (x >= b)
+  # is faster than findInterval(), which has dispatch overhead and uses binary
+  # search (overkill for 3-5 breakpoints)
+  if (length(x0) <= 10L) {
+    if (right)
+      function(x) { i <- 1L; for (b in x0) i <- i + (x > b); y[i] }
+    else
+      function(x) { i <- 1L; for (b in x0) i <- i + (x >= b); y[i] }
```
jdblischak, line 63: I found the construction of looping over `x0` to be counter-intuitive. Since the `x0` are ordered, I really wanted to find someway to get something like the following to work: `sum(x >= x0)`. But since this requires executing separately on each value of `x`, my solution ended up being rather convoluted too. Sharing just in case it could be informative:

```R
function(x) {y[vapply(x, function(z) 1 + sum(z >= x0), numeric(1))]}
```

jdblischak (approved):

R/utils.R
yihui: I also wish the loop could be vectorized but all ways that I could think of would be less efficient than the loop, e.g., the most elegant solution I could think of was:

```r
y[rowSums(outer(x, x0, '>=')) + 1]
```

but it was about 5x slower than the `for` loop.

yihui (commented):
@LittleBeannie I'm very confident that the new `stepfun2()` can safely replace `stats::stepfun()`, but please feel free to close the PR if you are not comfortable with changes in `expected_event()`.

R/utils.R
jdblischak: > ```r
> y[rowSums(outer(x, x0, '>=')) + 1]
> ```

I love that solution!
 
> but it was about 5x slower than the `for` loop.

What a shame. Unfortunately we have to go with the most performant solution for this critical code path.

LittleBeannie:
Thank you so much for your work on optimizing this. I really appreciate the effort and the improvements you've proposed.

The `expected_event()` function is a very low-level and complex component with many interconnected formulas. Because of its critical nature, our current policy is to only merge changes that have been scoped and suggested by Keaven, who is the primary architect of this module.

Therefore, I will be closing this PR for now. This is purely a process decision to maintain stability and is not a reflection on the quality of your work, which is excellent.

We can definitely re-open this or a similar PR in the future when the development roadmap for this module is updated.

Thank you again for your contribution!

PR #671, opened 2026-09-09
Assign a unique class to the output of gs_design/power_npe()

### LINKED ISSUE #657: Add a dedicated S3 class to gs_design_npe() output
## Motivation

`gsDesign2::gs_design_npe()` currently returns an object whose classes are only:

```r
c("tbl_df", "tbl", "data.frame")
```

The column schema is distinctive and useful (`analysis`, `bound`, `z`,
`probability`, `probability0`, `theta`, `info_frac`, `info`, `info0`, and
`info1`), but downstream packages cannot identify the object reliably without
matching its column names. In contrast, the other `gsDesign2` design
constructors return objects with a package-specific class.

This arose while adding direct support for `gs_design_npe()` results to
`gsDesignBroom::compare_designs()`:
https://github.com/keaven/gsDesignBroom/issues/2

## Suggested enhancement

Could `gs_design_npe()` add a dedicated S3 class while retaining its existing
tibble classes and behavior? For example:

```r
class(result) <- c("gs_design_npe", class(result))
```

This would allow downstream packages to register methods or dispatch to a
schema adapter using `inherits(x, "gs_design_npe")`, rather than relying on
heuristic column-signature detection.

## Possible acceptance criteria

- `inherits(gs_design_npe(...), "gs_design_npe")` is `TRUE`.
- The result continues to inherit from `tbl_df`, `tbl`, and `data.frame`.
- Existing printing, subsetting, and column names remain unchanged.
- The class is documented as a stable way for downstream packages to identify
  this result type.

This is not blocking the downstream implementation, but it would make the
integration more robust and easier to maintain.

jdblischak:
Close #657 

As discussed in group meeting, the output of `gs_design_npe()` and `gs_power_npe()` is too dissimilar to the output of the other group sequential design functions (eg `gs_design_ahr()`) to be assigned the existing class "gs_design", so I assigned them each a unique class to facilitate downstream S3 method dispatch.

Note that the existing documentation does not document the columns `probability0` returned by `gs_design_npe()`

```R
colnames(gs_design_npe())
## [1] "analysis"     "bound"        "z"            "probability"  "probability0" "theta"       
## [7] "info"         "info0"        "info1"        "info_frac"   
colnames(gs_power_npe())
## [1] "analysis"    "bound"       "z"           "probability" "theta"       "theta1"      "info_frac"  
## [8] "info"        "info0"       "info1"      
```

xref: motivating Issue (https://github.com/keaven/gsDesignBroom/issues/2), abandoned attempt to reuse class `"gs_design"` (https://github.com/jdblischak/gsDesign2/commit/62ffdf7fb32237390863eaa9257596cba23ce6dd)

### DIFF
--- a/tests/testit/test-developer-gs_power_npe.R
+++ b/tests/testit/test-developer-gs_power_npe.R
@@ -6,7 +6,7 @@ assert("The default of `gs_power_npe` is a single analysis with type I error con
     dplyr::filter(Bound == "Upper") |>
     dplyr::rename(analysis = Analysis, bound = Bound, z = Z, probability = Probability) |>
     dplyr::mutate(bound = tolower(bound))
-  (all.equal(x1, x2, check.attributes = FALSE))
+  (unclass(x1) %==% unclass(x2))
 })
 
 assert("fixed bound", {
@@ -29,7 +29,7 @@ assert("fixed bound", {
   ) |>
     dplyr::rename(analysis = Analysis, bound = Bound, z = Z, probability = Probability) |>
     dplyr::mutate(bound = tolower(bound))
-  (all.equal(x1, x2, check.attributes = FALSE))
+  (unclass(x1) %==% unclass(x2))
 })
 
 assert("Same fixed efficacy bounds, no futility bound (i.e., non-binding bound), null hypothesis", {
@@ -48,7 +48,7 @@ assert("Same fixed efficacy bounds, no futility bound (i.e., non-binding bound),
   ) |>
     dplyr::rename(analysis = Analysis, bound = Bound, z = Z, probability = Probability) |>
     dplyr::mutate(bound = tolower(bound))
-  (all.equal(x1, x2, check.attributes = FALSE))
+  (unclass(x1) %==% unclass(x2))
 })
 
 assert("Fixed bound with futility only at analysis 1; efficacy only at analyses 2, 3", {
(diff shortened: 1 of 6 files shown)


LittleBeannie (approved):

tests/testit/test-developer-gs_power_npe.R
```diff
@@ -6,7 +6,7 @@ assert("The default of `gs_power_npe` is a single analysis with type I error con
     dplyr::filter(Bound == "Upper") |>
     dplyr::rename(analysis = Analysis, bound = Bound, z = Z, probability = Probability) |>
     dplyr::mutate(bound = tolower(bound))
-  (x1 %==% as.data.frame(x2))
+  (all.equal(x1, x2, check.attributes = FALSE))
```
yihui: If they differ only in class, perhaps we can check for `unclass()`ed objects being strictly identical? That will be more rigorous since `all.equal()` tests for approximate instead of exact equality.

```suggestion
  (unclass(x1) %==% unclass(x2))
```

yihui (approved):
Just one small comment on tests.

keaven:
Looks fine to me.

jdblischak:
Two options to test this branch locally:

### 1. Install with `install_github()`

```R
remotes::install_github("jdblischak/gsDesign2@s3npe")
library("gsDesign2")
x1 <- gs_design_npe()
class(x1)
## [1] "gs_design_npe" "tbl_df"        "tbl"           "data.frame"
x2 <- gs_power_npe()
class(x2)
## [1] "gs_power_npe" "data.frame"
```

### 2. Checkout the branch from my fork

```sh
git remote add john https://github.com/jdblischak/gsDesign2.git
git fetch john
git checkout john/s3npe

# Use R to install local package, eg with `devtools::install()`

# Post-testing, return to main branch
git checkout main
```

tests/testit/test-developer-gs_power_npe.R
jdblischak: Great suggestion! Done in 4f58804b6e8777d905dbdde73be4a26789807f6b

PR #640, opened 2026-06-10
Add harm bound to AHR designs

LittleBeannie:
To solve issue #618.

@jdblischak: In addition to "Efficacy" and "Futility" bound, I added "Harm" bound, which sequentially leads to some changes in `gs_bound_summary()`. The changes in `gs_bound_summary()` is suggested by GPT5.5. Could you please review if these AI-suggested changes looks good to you?

PR #673, opened 2026-09-11
Retain harm bound in to_integer() for AHR designs

yihui:
## Problem

`to_integer.gs_design()` rebuilds an AHR group sequential design with `gs_power_ahr()` at the rounded integer event counts, but the reconstruction did not forward the harm bound arguments (`harm`, `hpar`, `test_harm`). As a result, the harm boundary was silently dropped whenever an AHR design with harm bounds was converted to integer events:

```r
x <- gs_design_ahr(
  analysis_time = c(18, 30),
  upper = gs_spending_bound,
  upar = list(sf = gsDesign::sfLDOF, total_spend = 0.025, param = NULL),
  lower = gs_spending_bound,
  lpar = list(sf = gsDesign::sfHSD, total_spend = 0.1, param = -2),
  harm = gs_spending_bound,
  hpar = list(sf = gsDesign::sfHSD, total_spend = 0.05, param = -2),
  test_harm = TRUE
)
unique(x$bound$bound)              #> "upper" "harm" "lower"
unique(to_integer(x)$bound$bound)  #> "upper" "lower"   <- harm lost
```

This surfaced while wiring harm bounds through the gsDesign2 Shiny app (keaven/gsDesign2Shiny#76), where integer sample sizes are on by default.

## Fix

Forward `harm`/`hpar`/`test_harm` from `x$input` into the `gs_power_ahr()` call in the AHR/WLR branch. These fields are always present on the input and default to `gs_b`/`-Inf`/`FALSE` for designs without a harm bound, so non-harm designs are unaffected. `gs_power_wlr()` does not accept harm arguments, so the change is guarded to the AHR case (`is_ahr`).

After the fix, `to_integer()` retains the harm bound and the cumulative harm crossing probability under H0 at the final analysis matches the input (0.05 in the example above).

## Tests

Added an independent test asserting the harm bound survives `to_integer()` and that the final cumulative H0 harm crossing probability equals the requested total spend. Existing `to_integer` tests still pass.

### DIFF
--- a/tests/testit/test-independent-to_integer.R
+++ b/tests/testit/test-independent-to_integer.R
@@ -215,24 +215,3 @@ assert("to_integer.gs_design performs correctly with large sample sizes", {
   (all(result$analysis$event %% 1 == 0)) # Ensure events are integers
   (all(result$analysis$n %% 1 == 0)) # Ensure sample sizes are integers
 })
-
-assert("to_integer.gs_design retains the harm bound (test.type 7/8 equivalent)", {
-  design_ahr <- gs_design_ahr(
-    analysis_time = c(18, 30),
-    upper = gs_spending_bound,
-    upar = list(sf = gsDesign::sfLDOF, total_spend = 0.025, param = NULL),
-    lower = gs_spending_bound,
-    lpar = list(sf = gsDesign::sfHSD, total_spend = 0.1, param = -2),
-    harm = gs_spending_bound,
-    hpar = list(sf = gsDesign::sfHSD, total_spend = 0.05, param = -2),
-    test_harm = TRUE
-  )
-
-  result <- to_integer(design_ahr)
-
-  # The harm bound must survive the conversion to integer events
-  ("harm" %in% result$bound$bound)
-  # Cumulative harm crossing under H0 at the final analysis matches the input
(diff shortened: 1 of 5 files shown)


tests/testit/test-independent-to_integer.R
```diff
@@ -215,3 +215,24 @@ assert("to_integer.gs_design performs correctly with large sample sizes", {
...
+  )
+
+  result <- to_integer(design_ahr)
+
+  # The harm bound must survive the conversion to integer events
+  ("harm" %in% result$bound$bound)
+  # Cumulative harm crossing under H0 at the final analysis matches the input
+  harm_h0 <- result$bound$probability0[result$bound$bound == "harm"]
+  (abs(max(harm_h0) - 0.05) < 1e-6)
+})
```
jdblischak: IIUC I think this test should be moved to `test-developer-to_integer.R`

jdblischak (commented):
xref: follow-up to https://github.com/Merck/gsDesign2/pull/640

tests/testit/test-independent-to_integer.R
yihui: Sure. Done. Thanks!

jdblischak (approved):

LittleBeannie (approved):

PR #674, opened 2026-09-11
Document developer vs independent test conventions

yihui:
Adds a **Test Categories** section to the `write-tests` skill documenting the `test-developer-*` vs `test-independent-*` file-naming convention, which was previously implicit.

- **`test-independent-*`** — validate outputs against an independent source of truth: a separate reference implementation (e.g. `gsDesign::gsSurv()`, `gsDesign::toInteger()`), a closed-form/hand-computed value, or a mathematical identity. Should not rely on gsDesign2 internals.
- **`test-developer-*`** — exercise the package's own implementation behavior: default arguments, S3 class/attribute plumbing (e.g. `uninteger_is_from`), internal helpers (`gsDesign2:::`), consistency between exported and internal variants (e.g. `gs_power_ahr()` vs `gs_power_ahr_()`), and "does this feature survive this transformation?" checks.

Motivated by the review on #673, where a `to_integer()` harm-bound test was initially placed in `test-independent-*` but belonged in `test-developer-*`. Documenting the rule so future tests land in the right file.

Docs-only; no code or test changes.

### DIFF
--- a/.claude/skills/write-tests.md
+++ b/.claude/skills/write-tests.md
@@ -23,6 +23,16 @@ tests/
     └── test-*.md           # Snapshot files (standalone, no .R file needed)
 ```
 
+## Test Categories: `developer` vs `independent`
+
+Every test file is named `test-developer-<topic>.R` or `test-independent-<topic>.R`. The prefix signals **how** the test establishes correctness — pick the right one so tests stay where reviewers expect them.
+
+- **`test-independent-*`** — validate outputs against an **independent source of truth**: a separate reference implementation (e.g. `gsDesign::gsSurv()`, `gsDesign::toInteger()`), a closed-form/hand-computed value, or a mathematical identity. These answer "does the package produce the statistically correct number?" and should not rely on gsDesign2 internals.
+
+- **`test-developer-*`** — exercise the package's **own implementation behavior**: default arguments, S3 class/attribute plumbing (e.g. `uninteger_is_from`), internal helpers (`gsDesign2:::`), consistency between exported and internal variants (e.g. `gs_power_ahr()` vs `gs_power_ahr_()`), argument handling, and "does this feature survive this transformation?" checks. These answer "does the code behave as engineered?" and may inspect internals freely.
+
+Rule of thumb: if the expected value comes from *outside* gsDesign2, it's `independent`; if it comes from gsDesign2 itself or from how the function is wired, it's `developer`. Add a new test to the existing file for its topic and category (e.g. a `to_integer()` behavior test goes in `test-developer-to_integer.R`, next to the other harm/attribute tests).
+
 ## Core Pattern
 
 ```r


jdblischak (commented):
@yihui thanks for opening this Issue. I agree that these testing conventions should be clarified and documented, since I am not 100% sure how this is supposed to work.

Here is a what I think I know. As part of qualifying analytical software to be used in regulatory submissions, there can be a "double programming" step where an independent developer writes validation tests. In other words, the independent tester writes their own code to implement a particular method, and then tests to confirm that the package software returns the same result. These tests are in the `test-independent-*.R` files.

Here is the part I don't know. Can we as the developers of the software ever write "independent" tests? What if we write a test that uses a mathematical identity? Should these go in "independent", or should by definition any test written by a developer go in `test-developer-*.R` (since we can't be independent of ourselves, and any misunderstanding we have when writing a function will be carried over into our testing of the method, even if we use an orthogonal computation technique).

yihui:
@jdblischak Thanks for sharing your knowledge about these tests! I guess @LittleBeannie knows more than us and can help clarify.

jdblischak:
> Can we as the developers of the software ever write "independent" tests?

To follow up on this, my rule of thumb has always been to focus on _who_ wrote the tests, not _what_ is being tested or _how_. 

The benefit of this approach is that the rule is super simple to apply: unless you were explicitly requested to write independent tests to validate the statistical software, then the tests you write should be labeled as "developer" (or "unvalidated").

The obvious downside is that this blunt approach is not very informative about the nature of the tests. For example, `test-independent-check_arg.R` contains tests written by an independent tester that confirm known errors are caught, but in this case there is no independent implementation, ie double programming.

Also, it is unclear to me the rules of updating tests in the `independent` test files. At times we purposefully update the statistical behavior of our code, which requires us to update the independent tests accordingly. How much can we alter these tests before they no longer count as an independent validation of our software?

jdblischak:
Decision from internal discussion: Tests should be saved in `test-independent-` if 1) they were written by an independent reviewer, and/or 2) the test uses an independent implementation.

yihui:
@jdblischak Thanks! I've updated the PR accordingly.

.claude/skills/write-tests.md
```diff
@@ -23,6 +23,23 @@ tests/
...
+2. **It uses an independent implementation** — the expected value comes from *outside* gsDesign2: a separate reference implementation (e.g. `gsDesign::gsSurv()`, `gsDesign::toInteger()`), a closed-form/hand-computed value, or a mathematical identity.
+
+Everything else is **`test-developer-*`** — tests the developer writes against the package's own implementation behavior: default arguments, S3 class/attribute plumbing (e.g. `uninteger_is_from`), internal helpers (`gsDesign2:::`), consistency between exported and internal variants (e.g. `gs_power_ahr()` vs `gs_power_ahr_()`), argument handling, and "does this feature survive this transformation?" checks. These may inspect internals freely.
+
+Notes:
+
+- The primary signal is **who** wrote the test, not what or how it tests. Unless you were explicitly asked to write an independent validation, label your tests `developer`. (An independent reviewer may still write `test-independent-*` tests with no double programming — e.g. `test-independent-check_arg.R` confirms known errors are caught, with no reference implementation.)
```
jdblischak:
```suggestion
- The primary signal is **who** wrote the test, not what or how it tests. Unless you were explicitly asked to write an independent validation, you should most likely label your tests `developer`. (An independent reviewer may still write `test-independent-*` tests with no double programming — e.g. `test-independent-check_arg.R` confirms known errors are caught, with no reference implementation.)
```

Truly just a suggestion. I was trying to come up with a way to soften this statement without being overly verbose.

We want to leave open the option that a developer may write an "independent" test if they go through the trouble of writing an independent implementation, while recognizing that this is rare, and that a developer should default to writing their tests in `test-developer-` files.

jdblischak (approved):
Thanks @yihui for starting this discussion and improving our documentation!

.claude/skills/write-tests.md
yihui: Makes sense. Thanks!

